PR #5, opened 2017-12-03
Modifications to the Nuclear Blast Algorithm

grangerave:
First- really big fan of the mod!

I've proposed several changes to the way the nuclear blast algorithm works. Some of the things that I thought could be improved included:
-the slow spread speed of the destruction
-the limited interaction with other mod's blocks
-the fixed nature of the explosion shape
-the fact that blocks would be destroyed even if surrounded by blast resistant blocks

So I've made some changes for your consideration!
**First, the blast shape.** 

(Somewhat) realistically, you expect a blast in the air to have a spherical shape, while a blast on the surface would have the bottom radius reduced by the material below.
Here is the previous explosion shape:
image
The central area, called by `destruction()` shows this behavior, but the `waste()` and `vapor()` areas are simply spheres, leading to a disproportionate level of destruction below ground (and a good number of extra block calls).

With the changes, at the time when the explosion entity is created, the bottom radii of all 3 zones are determined based on the blast height, as shown below (ideally, the top radii would attenuate the farther below ground the blast is, but given how thin the crust is anyways, this isn't noticeable enough to implement).
image

This results in the following picture: when the blast is detonated far below ground, the lower half of the sphere is compressed (much like the previous code). _(note: explosion radius reduced to 70 to make effect more visible )_
image
If we detonate at the surface, the shape is still the same:
image
But if we detonate the blast in the air, the shape starts to shift to a full sphere:
image
image
This means aerial blasts are destructive too! (much like IRL)

**Second, the blast code itself**
Specifically `destruction()`, `vaporDest()` and `wasteDest()`. Instead of a long if statement that checks for specific block instances, we cache the block, then check its blast resistance to decide whether to destroy it or not. 

`Block b = world.getBlock(x,y,z);`
`if (b.getExplosionResistance(null)>=200f) {	//500 is the resistance of liquids`
`	...`

This means the explosion won't destroy other mods' blast resistant blocks (compatibility woo!), but I've had to up the blast resistance of `waste_log` and `waste_planks`.  The cutoff is set to a blast resistance of 1000f _(note: getExplosionResistance() returns 1/5 of the value for some reason)_, which is well above liquid blocks, but includes things like obsidian, end portals, etc. If the block is not one of the blast bricks from the mod, there is a chance for the block to be turned into `Block_scrap` (the higher the resistance, the lower the chance).

The `vaporDest()` function is modified in a similar way. We remove the following: 
-Blocks with resistance below 2.5 (dirt is 2.5)
-ALL Liquids
-Non-solid blocks below 15, excluding chests and farmland
This gives roughly the same functionality as before, but destroys more things that are usually destroyed!

The `wasteDest()` function is mostly the same, with the only addition being that solid, wooden blocks are turned into `waste_planks`.

Also, all the `SetBlock()` calls are switched to not trigger the neighors, in order to minimize ticking.

**Finally, a crude attempt to model Blast Shielding**
In an ideal world, we'd use some sort of raytracing to model the blast attenuation, which would be slowed down by a blast resistant block. But given the enormous nature of these explosions, this would be CPU-meltingly impossible! So I've implemented a compromise:
-Assume blast proof structures are roughly cubical
-If we hit a block below the epicenter thats not destroyed/broken by the blast:
-Skip the next few blocks below it
The higher the block's explosion resistance, the more blocks we skip. This results in a crude model of blast attenuation! Less floating structures once the blast is done, and paintings/beds/flowerpots inside bunkers will stay untouched! (Wood is still charred, because why not).

I'm running all this on a potato, but it looks like most of these changes have improved the total time from start to end of an explosion (for example, a full Littleboy blast finished in less than a minute, compared to several minutes before). And most importantly, the blast speed is now a setting in the config file, so beastly setups can run the detonations faster!

Thank you for the consideration, and good luck with development!

### DIFF
--- a/com/hbm/blocks/ModBlocks.java
+++ b/com/hbm/blocks/ModBlocks.java
@@ -20,6 +20,7 @@ import net.minecraft.block.material.MapColor;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.material.MaterialLiquid;
 import net.minecraft.creativetab.CreativeTabs;
+import net.minecraft.init.Blocks;
 import net.minecraftforge.fluids.Fluid;
 import net.minecraftforge.fluids.FluidRegistry;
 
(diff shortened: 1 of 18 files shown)


HbmMods:
Woah! I didn't think somebody would do a PR on this project, let alone one that changes such a fundamental and complicated part! Recently I considered redoing the algorithm to function more like the default minecraft one but still works in steps so it would be usable for these enormous scales, but the one you made even keeps the round shape I like so much.

You, sir, did an amazing job!

PR #45, opened 2020-04-27
Ported a bunch of guns and fixed bugs with the gun stystem and introduced new features to it

Syncinus:
Might need some tweaking but has been tested and shouldn't have any breaking bugs.


HbmMods:
There's quite a few things I wanted to do too (like using the dart renderer for all newer laser weapons) but also a few changes i don't *quite* agree with. Good job anyway, I'll get back to that once I overhaul the remaining weapons which should be soon-ish.

PR #65, opened 2020-06-23
Old style Eclipse workspace generation via `./gradlew eclipse`

OvermindDL1:
This 'should' be everything to generate the old style Eclipse workspace.  I cannot really attest to if it works or not as it was always tested by others, but they did confirm it worked fine at least up until a few years ago (most have changed to IntelliJ since).

Just run `./gradlew eclipse` to generate the `eclipse` folder in the project, then open that folder as a normal eclipse-style project/workspace.  If something about it doesn't work then tell me and I can update this PR before you merge it.


OvermindDL1:
Oh, I also left in `options.compilerArgs << "-Xlint:deprecation" << "-Xlint:unchecked"` by accident in the build.gradle file, just noticed, sorry...  Those will show warnings that were previously hidden when you build.  You can remove that line if you don't want it, but I do recommend fixing those warnings in the code and leaving that line enabled instead.  It will still build even with the warnings so no change there.

HbmMods:
It appears as if pulling this commit atomized my eclipse folder, even though it was in .gitignore, and now Eclipse is acting up too. I suppose it might be time to switch to IntelliJ afterall.

OvermindDL1:
Pulling the commit wouldn't touch an existing `eclipse` folder.  Only thing that could have would have been something like locally running `git clean --force -xd` or `git reset --hard` or something like that.  Running the command `./gradlew eclipse` would have rebuilt the eclipse folder however.

How is Eclipse acting up though I'm curious?

HbmMods:
The `eclipse` folder doesn't have all the game stuff in it, instead a `.metadata` folder and a symlink called `Minecraft` which leads back to the project's root directory. That might explain why eclipse doesn't want to start at all.

OvermindDL1:
I'm not really sure, I didn't build the overall setup for it, only the templating for it...  ^.^;

Got it working though?  What changes fixed it so I can propagate them back elsewhere?

HbmMods:
I took a look at the build script and from what I can tell, it's supposed to link to the root directory, but I have no idea why exactly anybody would want that. The only way I could fix it right now would be by reverting the workspace to what it was before, but then I'd either have to use the older NEI libraries or ask you for the jars and add them manually ¯\\\_(ツ)_/¯

OvermindDL1:
The jars are available on maven, some ID but leave off the `:dev` when downloading.

By symlink are you referring to the line of `ant.symlink(resource: "../", link: "eclipse/Minecraft")` though?  It's symlinking the `eclipse/Minecraft` directory to the projects directory so the directories can link properly.  That shouldn't do anything to the eclipse project itself as it shouldn't be touching the `Minecraft` symlink at all, it only exists so when actually running minecraft inside the IDE it knows how to access it's instance directory again (since eclipse sets the working directory to 'eclipse' for some reason and couldn't figure out how to go 'up' I was told).  I'm really no fan of eclipse, lol...  ^.^;

PR #85, opened 2020-08-13
Update zh_CN.lang

yeyuluo123:


HbmMods:
The branch looks like it's conflicting, I'll merge it locally. Are you still using an older workspace where the project root is still /src/?

PR #146, opened 2021-03-13
config file to build.gradle

pbk20191:
this will help update and building new workspace. from now on you dont need to download extra files like codechicken and nei

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,10 @@ allprojects {
     sourceCompatibility = 1.8
     targetCompatibility = 1.8
 }
-
+compileJava.options.encoding = 'UTF-8'
+tasks.withType(JavaCompile){
+	options.encoding = 'UTF-8'
+}
 file "build.properties" withReader {
     def prop = new Properties()
     prop.load(it)
@@ -40,6 +43,15 @@ minecraft {
     version = "${config.minecraft.version}-${config.forge.version}-${config.minecraft.version}"
     runDir = "eclipse"
 }
+// configure  the source folders
+sourceSets {
+    api {
+		java {
+			srcDir 'api'
+		}
+    }
+}
+
 repositories {
     maven {
         name 'Forge' // Applied Energistics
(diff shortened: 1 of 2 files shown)


HbmMods:
How exactly does it work? And why does it have so many maven repositories of other mods in it?

build.gradle
```diff
@@ -1,56 +1,101 @@
...
+archivesBaseName = "hbm"
+
+minecraft {
+    version = "${config.minecraft.version}-${config.forge.version}-${config.minecraft.version}"
+    runDir = "eclipse"
+}
+repositories {
+    maven {
+        name 'Forge' // Applied Energistics
+        url 'http://files.minecraftforge.net/maven'
```
pbk20191, line 58: this contains AE2 and other mods

build.gradle
```diff
@@ -1,56 +1,101 @@
...
+    }
+    maven {
+        name = "chickenbones"
+        url = "http://chickenbones.net/maven/"
```
pbk20191, line 62: this is for chickenbones: NEI, codechickenCore, CodechickenLib

pbk20191:
> How exactly does it work? And why does it have so many maven repositories of other mods in it?

most of this maven are for just for future dependencies, but only needed is name = "chickenbones", you can remove other maven. I would recommand you to don't remove at least <name "Mobius Repo"> maven for Waila support.

build.gradle
```diff
@@ -1,56 +1,101 @@
...
+
+apply plugin: 'forge'
+
+allprojects {
+    apply plugin: 'java'
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+}
+
+file "build.properties" withReader {
```
pbk20191, line 32: this will read "build.properties" and extract data

pbk20191:
> How exactly does it work? And why does it have so many maven repositories of other mods in it?

before doing any gradle command, it will read build.properties file and check dependancy. and if dependancy fails than it will automatically download and complie the required files

pbk20191:
umm, this is out of topic. Could you please move /asset into the /resource/ directory? putting asset file inside the source file can cause problem with other resource loading mods. and that is not friendly for modders

pbk20191:
sorry.. I accidently mixed up my PR. I will remake the PR when I'm done.

PR #147, opened 2021-03-13
source and mod file management

pbk20191:
mcmod.info can now managed by build.properties

gradle is updated to 4.4.1

gradle will now automatically fetch the required dependency

api codes are separated for modding Convenience.

build.properties now handles mod, forge, minecraft, and required
dependcy mods version

assets should be out of the source. If not, it gives huge inconvenience
to modders. Also it might cause problems with other resource loaders

.p.s. I think you are committing code quite often. So I suggest you to create /dev branch as a test version of master branch.  Dev branch will protect your files from Harmful PR, which will damage you source. (maybe this branch is harmful PR)


HbmMods:
I'm afraid I can't merge this PR, for one because I can't review 5k files, but also because I'd much rather do structural changes like these myself, bit by bit.

What interests me more though is the changes to the build system, I don't actually understand how gradle or maven work, so I'm wondering what changes are required on my part. I remember PR #65 also changing the build system which caused it to be unusable with eclipse and I ended up having to delete cached gradle files since reverting the changes would still leave the project broken.

pbk20191:
> I'm afraid I can't merge this PR, for one because I can't review 5k files, but also because I'd much rather do structural changes like these myself, bit by bit.
> 

sure.  I expected that, this PR won't be merged. cause of HUGE change. I'm satisfied with just accepting the concept of my PR. check my fork. "dev" branch. try my branch and reply to me if anything else go wrong or need more infomation

pbk20191:
> What interests me more though is the changes to the build system, I don't actually understand how gradle or maven work, so I'm wondering what changes are required on my part. I remember PR #65 also changing the build system which caused it to be unusable with eclipse and I ended up having to delete cached gradle files since reverting the changes would still leave the project broken.

This changes is just controlling the versioning system. and automatically fetch the required dependency from maven. gradle will read the build.properties ( minecraft.version, forge.version, mod.version, required dependency.version), then gradle will use those data for excuting gradle task. Also "gradle build" will automatically change the "mcmod.info" with correct version.

API parts, is just added for more convenience. It is useful if it is changed, but it is okay not to do so. I seperated the API parts from the source. Gradle will link the api to the source when "gradle eclipse" is excuted. And Gradle will compile the api and merge the source and api together when "gradle build" is excuted.

HbmMods:
Alright I've moved the assets, later this week I'll probably experiment with your proposed gradle script changes. As for the dev branch, it does sound like an idea worth considering, we'll see.

I'm leaving this PR open for now until all the proposed changes have been made. Thank you for your interest in the project and your contributions.

pbk20191:
about commit: 9797fcc
I updated the PR, to the recent version of master branch. Also removed asset changement for preventing confliction.

HbmMods:
Closing due to being resolved in PR #268 .

PR #196, opened 2021-06-04
Energy API finish

eepy-goddess:
Hi people reading this pull request. I noticed the IEnergyConsumer and IEnergySource parts of the hbm api haven't been finished yet, and the last commit to these interfaces were 10 months ago, so I don't know if y'all forgot about it. I deleted the ISource and IConsumer interfaces and copied their code to the IEnergyConsumer and IEnergySource in the hbm api section of the source and I replaced all usages of ISource and IConsumer to IEnergySource and IEnergyConsumer. I also added an HBM energy util class that has a static method to perform the ffgeua on tile entities in seperate mods without needing the HBM Library class.


HbmMods:
I scrapped the energy API because the current wire system is so horrible that I do not want anyone to use it. As a matter of fact I was planning on reworking the system to something more sensible in the upcoming week since that will allow me to make more and less laggy cable variants more easily. The new system would ideally have a comprehensive API which would work even when implemented in a standalone mod without NTM, similar to the COFH redstone flux API.

In the unlikely event that somebody already used the current interfaces for mod compat (the ones from the `com.hbm.interfaces` package, since the actual API is unused) I'm planning on keeping those compatible, at least the `IConsumer`.

PR #228, opened 2021-07-05
Finally

UFFR:
satori dead inside


HbmMods:
image

I will thank you in the distant year of 35,924 when I'm done reviewing and merging all these changes.

UFFR:
In all seriousness, there are still a few quirks here and there that aren't completely worked out yet that I either just couldn't find or don't know how to fix. Most of them are pretty simple so should be resolved easily. As for machines, items, or other features that are beta stage (which most of them are), I guess you'll either have to quickly finish or disable them. The RTG blast furnace is pretty much done though, the Watz Super-compressor works, but only with "shaped" recipes and it has no model, the Tsukuyomi weapon works, but is unfinished and unbalanced, the TWR is still a placeholder, the divine light weapon only works in flood mode since I couldn't figure out rays, the atomic clock _sort of_ works, but the time system behind it is still a bit new and I don't have any debugging commands for it, the new guns work fine, their animations are rather unsophisticated, but are passable, the energy gun base class seems to work as well as the fast chargers for the moment. Custom cores and warheads are kind of in, but they aren't functional and likely won't be for a while. There may still be a few missing recipes either due to me missing them or simply not having the resources to add them, such as the Niobium superalloy recipe, which requires that alloy forge thing proposed a while back.

UFFR:
There are also temporary/placeholder recycling recipes for some depleted fuel types (Schrabidium fuel) but you can change/remove those entirely at your discretion.

PR #239, opened 2021-07-24
Bullet Press Stamps are Shredder Blades

MartinTheDragon:
Whilst implementing Shredder Blades in the 1.16.5 version, I found that, in fact, bullet press stamps are shredder blades, and will also work as such.

I did not create a build of this version, but I only changed four lines from `ItemBlades` to `ItemStamp`.


HbmMods:
The press spaghetti code is haunting me to this day

PR #262, opened 2021-08-24
Fix Glowing Mushrooms appearing in water

MartinTheDragon:
With the Glowing Mycelium config option enabled, it would spread underwater and start spawning Glowing Mushrooms. The mushrooms would get in turn broken by the water, thus creating infinite Glowing Mushrooms.

Both things have been fixed. Glowing Mycelium does not propagate underwater anymore and no longer spawns Glowing Mushrooms if the block above is not air. Glowing Mycelium will still remain underwater though, because of the light it emits, and because of 1.7 logic, making it totally fine even for Grass Blocks to stay if there is enough light.

Hilariously, before releasing 1.16.5-0.0.8.0, my code had a similar issue where the Glowing Mushrooms would appear in flowing water. Anyway, I should really go to bed now...

P.S. The .gitignore file had two entries with /.gradle/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,7 @@
 /eclipse/
 /lib/
 /.git/
-/.gradle/
 /.settings/
 /.classpath
 /.project
-
+.idea/
--- a/src/main/java/com/hbm/blocks/generic/WasteEarth.java
+++ b/src/main/java/com/hbm/blocks/generic/WasteEarth.java
@@ -1,13 +1,9 @@
 package com.hbm.blocks.generic;
 
-import java.util.Random;
-
 import com.hbm.blocks.ModBlocks;
 import com.hbm.config.GeneralConfig;
 import com.hbm.lib.RefStrings;
-import com.hbm.main.MainRegistry;
 import com.hbm.potion.HbmPotion;
-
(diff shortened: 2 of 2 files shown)


9Valjew:
snitch

HbmMods:
Honestly I should move spreading behavior to the mushroom itself instead of the grass blocks, since that way the mushrooms can spread onto other thing that aren't specifically those blocks and it would remove a lot of awful clutter in the dead grass class. Haven't taken a look at the mushroom class yet, but knowing me it most certainly looks like a car crash, so I might as well redo all of that stuff too.

HbmMods:
I haven't actually tested the changes yet, I'll do that when i get back home. The changes are:

- Glowing mushrooms now spread on their own, the code is near identical to vanilla mushrooms
- There's now a slightly larger list of blocks that mushrooms can actively spread to
- Mushrooms can now be placed on blocks like stone where vanilla mushrooms can be placed too
- Dead grass and glowing mycelium now turn vanilla mushrooms standing on top of them into glowing mushrooms. "Natural" glowing mushrooms after nuclear explosions are much less likely now because instead of a vanilla shroom having to just be near dead grass, it has to spread onto the grass before a glowing shroom can grow.

PR #267, opened 2021-09-01
Autonomized build system

Enginecrafter77:
The build system should now be fully autonomous. That means that the build should run just fine without any user intervention upon cloning the repository. Additionally, the README was updated with new instructions for building the mod and setting up modding workspace.

Notable changes:
 * Replaced local dependencies with maven ones (https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/commit/9a41ccc0ba8c51334bda9c8acc0a598f0511bf45)
 * Updated .gitignore (https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/commit/9a41ccc0ba8c51334bda9c8acc0a598f0511bf45)
 * Added updated building instructions (https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/commit/6c63313842f85f6e6c86e10214c2de910c4ff9d2)
 * Added automatic project versioning (https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/commit/4e641e8e7ce7a31b1a1e530971a8c81e68c4ec82)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,22 @@
-/.gradle/
-/bin/
-/build/
-/eclipse/
-/lib/
-/.git/
-/.gradle/
-/.settings/
-/.classpath
-/.project
+# eclipse
+eclipse
+bin
+*.launch
+.settings
+.metadata
+.classpath
+.project
 
+# idea
+out
+*.ipr
+*.iws
+*.iml
+.idea
+
+# gradle
+build
+.gradle
+
+# other
+run
--- a/README.md
+++ b/README.md
@@ -4,39 +4,53 @@ https://minecraft.curseforge.com/projects/hbms-nuclear-tech-mod?gameCategorySlug
 
 For a 1.12 fork, check this link: https://github.com/Drillgon200/Hbm-s-Nuclear-Tech-GIT/releases
 
-## Installation Instructions
+## Building from source
 
 Tired of waiting until the next version comes out? Here is a tutorial on how to compile the very newest version yourself:
-1. Download minecraft forge 1.7.10 src
-2. Unpack it somewhere
-3. Download the source and insert it into the same folder. This will overwrite a couple of files, for example build.gradle
-4. Run `gradlew setupDecompWorkspace` on windows or `./gradlew setupDecompWorkspace` on linux (remember: do not take the name of root in vain). Running this command on a blank forge workspace will yield errors, so it's important you insert the mod's source in step 3. The source has corrected versions of guild.gradle and the gradle wrapper, if done correctly it should run just fine.
-5. The `build.gradle` file will reference these following files:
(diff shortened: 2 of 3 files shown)


HbmMods:
NEI, CCC and CCL download automatically now which is good, but it seems like it ends up somewhere in the `.gradle` folder. Is there a way to automatically add it to the project JARs, similar to how forge sets them up? If not then I'll have to add a notice to the `README`, since adding the jar files manually requires knowing where they ended up.

Additionally, is there a way to automatically read the version from the `mcmod.info` file? The `com.palantir.git-version` plugin depends on git being installed, and I doubt it will reproduce my versioning instead of just stamping the latest commit ID onto the file name. If not, then I could probably cobble something together with the `build.properties` file from PR #146

HbmMods:
Alright, I got the versioning right by using a `JsonSlurper` in the build script. It still includes square brackets which I can't seem to find the source of, but that's good enough for now.

Enginecrafter77:
> NEI, CCC and CCL download automatically now which is good, but it seems like it ends up somewhere in the `.gradle` folder. Is there a way to automatically add it to the project JARs, similar to how forge sets them up? If not then I'll have to add a notice to the `README`, since adding the jar files manually requires knowing where they ended up.

Why would you need the jars? I mean, you can find them somewhere in `$HOME/.gradle/caches/...` but why would anyone need that?

----------------

And besides, I added the git versioning as an optional feature, feel free to disable it if you want. I use it because I personally dislike repeated commits like "Bumped version to blah blah blah", which only change the version number. What the git version plugin I included does is that when you release a new version, a git tag is added. Then, the plugin detects that tag and puts the tag name as version. For example, you release version 1.2.0 (HEAD is tagged with tag 1.2.0), then the git version plugin uses that as the version number. IF the HEAD is **not** tagged, then it uses format like `[nearest tag]-[distance]-[commit]`, which may be useful for development versions. There is also a better plugin but it unfortunately doesn't work with the ancient gradle versions ForgeGradle 1.2 requires.

So, the "version bumping workflow" (bump version, tag version, create github release, upload binaries), becomes:
 1. create release on github
 2. git pull (pulls new tags)
 3. ./gradlew build
 4. upload binaries

HbmMods:
> Why would you need the jars?

For just building the project, having the jars buried somewhere for the compiler to use is fine, but when someone actually wants to make changes then it's necessary to add those libraries to the project properties through the IDE. In addition, the downloaded libraries only include the `dev` versions without the `src`, making it impossible to actually read the libraries from within the IDE without running a decompiler.

Enginecrafter77:
~~Actually, by using maven the sources are almost always visible. Most of the time maven repositories also include the source. And the IDE can directly see those sources. So if you CTRL+click some class from library it should be visible. And you don't even need to decompile anything (decompilation is messy anyways).~~

Oh wait my bad the system doens't download the sources. Idk why I will try to fix that. Codechicken's maven it really messed up.

PR #268, opened 2021-09-01
Fixed messy codechicken maven libraries

Enginecrafter77:
The codechicken maven libraries are now properly downloading sources along with the dev jars. Those source jars are automatically attached to the development jars, so you can see them in eclipse.

This pull request addresses follow-up conversation in https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/pull/267.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,32 @@ minecraft {
 	runDir = "eclipse"
 }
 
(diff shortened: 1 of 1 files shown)


HbmMods:
Thank you! The sources seem to work perfectly, and the libraries are attached automatically after running `./gradlew eclipse` after building.

PR #303, opened 2021-10-10
added anvil recipes

Vaern:
here you go bob


HbmMods:
Merged locally. I lifted some of the 528 mode penalties to reduce clutter + I don't think whether you get back your lead plates from a barrel of nuclear waste or not makes that big of a difference.

flop

Vaern:
Fair enough lmao

PR #315, opened 2021-10-18
Reworked radiation math with the reacher equipped

TehTemmie:
This change makes reachers reduce incoming radiation that is under 1 RAD/tick.


HbmMods:
Assuming that I've interpreted the function correctly
image
wouldn't that mean that the effective dose would never exceed 0.207 RAD/t? At least that's where the function peaked on the Desmos graphing calculator.

TehTemmie:
This is what the function is supposed to be
image
I really hope I didn't mess up my parentheses

HbmMods:
Oops, my bad

HbmMods:
grafik

PR #322, opened 2021-10-29
Make marking pipelines easier and quicker

MartinTheDragon:
Shift-right-clicking pipes with fluid identifiers will now change connected pipes within a maximum distance of 64 blocks.

You may want to add some info text in the fluid identifier's tooltip.
Or make the maximum distance configurable.


MartinTheDragon:
Added info text

HbmMods:
heythere

MartinTheDragon:
Ye I just looked at the PR and you merged it right before my eyes

HbmMods:
m a g i c

PR #326, opened 2021-11-01
helium things

Doctor17-git:
ok since that my first PR (expect .lang files), there may be some fucky wucky with files, because i using gitkraken in first time


HbmMods:
Merged locally. From what I've seen your PR doesn't actually add a delivery rocket, I took the liberty of simply adding another subtype to the existing delivery rocket and a second loot pool for that type, consisting of mostly moon turf and small parts of iron and lithium.

Doctor17-git:
yeah thanks

PR #339, opened 2021-11-22
ZIRNOX bug fixes

Vaern:
I (hopefully) fixed the issue with rounding for the water consumption/steam production along with the depletion of certain slots in the ZIRNOX.
Manually merging this latest commit should be fine


HbmMods:
Merged locally.

PR #363, opened 2021-12-07
Fixed rounding errors with large reactor

TehTemmie:
Large reactor rods only affected heat produced (which in turn affects steam production) in increments of 10% (for example, 19% control rod level produced the same heat as 10%) while still consuming more fuel. This change fixes that error, assuming it wasn't intentional. (Yes, I suck at pull requests)


HbmMods:
I know some people already claimed this issue to exist but in all these years I never even noticed anything. In any case it most definitely wasn't intentional.

HbmMods:
Perhaps I should fix the known rounding errors for steam conversion now since that would prevent the water cycle from being lossy which is kind of a big deal for people who want closed loop water cycles.

PR #365, opened 2021-12-12
Update zh_CN.lang + a few fixes for en_US.lang

TehTemmie:
Why did I spend this much time on the Chinese localization, why


HbmMods:
I received a new chinese localization file about an hour ago, so I don't think i need to merge this now (not that I could), as for the missing english one, I added that manually.

PR #387, opened 2022-01-12
Pu238be plate fuel, lead-200 changed to lead-209

Vaern:
pepehands


Vaern:
you don't play game so they are easy do you

PR #396, opened 2022-01-20
reacher funnies

Vaern:
make sure to make sure that the rubber gloves will properly render on the player


TehTemmie:
Shouldn't the radiation functions for 528 and normal mode be swapped? 528 is meant to be "hard mode", but the 528 function approximates a square root, which grows much more slowly than a linear function, which is what normal mode uses. Granted, the 528 function grows more quickly at first, but after ~2400 RAD per tick, the linear function becomes greater than the 528 function.

Vaern:
You're completely right, let me fix that

HbmMods:
grafik

Assuming I've interpreted this function correctly, what would happen if the original radiation were <0.24?

HbmMods:
nvm I just fucked up the brackets again, it's the same function as in #315 (where I screwed up the math as well)

HbmMods:
grafik

Issue #336, opened 2021-11-15
Reminder about the Deuterium tower funnies + list of my sins


7pheonix:
I never got time to commit this, so this is partly my fault. The deuterium tower's anvil recipe should use a tier 4 anvil, not a tier 5. The Deuterium tower is also missing a tile mapping, which is also my fault. Finally in the list of things i fucked up, the line `if(age >= 2)` in TileEntityDeuteriumExtractor, the 2 should be a 10 to limit deuterium extraction speeds, however you may want to adjust this for balancing purposes if enough people complain about it being slow.

### Please solve this equation to x

x²+3x-19=√(9x²)+6

x = 8.83
x = -2.83

UFFR:
I believe your algebra is flawed.

-- 
UFFR

HbmMods:
I just noticed that the deuterium extractor uses both the old `AuxElectricityPacket` and new `NBTPacket` to sync the energy level to the client, even though I can't seem to find any instance where that information is used. Could I safely remove that?

PR #406, opened 2022-01-28
flashlead

Vaern:
:reimumunch:


HbmMods:
So what's the big idea behind Ghiorsium? From the recipes it looks like the gating is identical to virgin balefire but with the added Pb-209 (=Australium) requirement. Did you have any other uses in mind besides armor mods?

Vaern:
The idea is that Ghiorsium-336 will serve as a stepping stone for the transschrabidics, whether that be via direct use in a breeding reactor or depletion in the CANDU or RBMK in conjunction with a fissile element. Pheo's already planning out the fissile isotope of Ghiorsium for schrabidium recycling; the flashlead is to serve as a post-fusion way to mass produce Gh-336.

HbmMods:
grafik

PR #420, opened 2022-02-02
I can do anything

Vaern:
jevil


HbmMods:
I threw out the registering code for radiolysis cracking recipes and replaced it with one that synchronizes the recipes from the refinery recipes (since I'll probably add more cracking recipes soon and all the recipes are supposed to be the same anyway, I think).

Hydrogen displaying as lava was due to the wrong texture sheet index being used (on that note, the texture sheet system is terrible and the new fluid system will get rid of it soon anyway).

As for the NPE in HazardTypeHot, while it kinda looks terrifying to me it's good that seemingly nobody has run into it yet.

Vaern:
Yeah, makes sense. I reckon keeping the quartet instead of changing it to fluid stacks is a matter of "if it ain't broke, don't fix it," anyway. The tank texture was an oversight on my part.

As for the NPE, I remember combing over the code and yet I still missed it.
You haven't been on the server so I can see why you'd think that, but a few people have indeed ran into it.

HbmMods:
So hot items _do_ crash instantly? And nobody opened an issue?

Vaern:
Yeah uh, for the first time in this mod's life people have confined the issue to crash-reports

PR #493, opened 2022-03-09
Add serialization for EntityFallingNuke

MartinTheDragon:
When unloading the chunk that has a falling custom nuke in it, coming back to it would not retain the explosion values.


HbmMods:
grafik

PR #494, opened 2022-03-09
Backport the faster fallout algorithm

MartinTheDragon:
No redundant fallout calculations anymore. Block transformation chances still need to be adjusted though. You can either do that by pushing to this branch or giving me the task. For now, this will be a draft.
There is also a TODO in EntityFalloutRain that calls for self-adjusting fallout chunk precision.


MartinTheDragon:
The TODO is finished, the precision is self-adjusting now.
But I have noticed the algorithm heavily impacting server performance in 1.7.10 for some reason (barely noticeable in 1.18 though), with an average of just 8 TPS, also slowing down the `EntityFalloutRain` tick speed. This *may* be ignored, but perhaps could be remedied by storing a cached version of the current chunk and letting the `stomp` method use that?

MartinTheDragon:
Fallout chunk computations now have four ticks of delay in between, which seems to go way easier on the server, maintaining 20 TPS.

I think I'm done, lezgo!

HbmMods:
flop

PR #546, opened 2022-04-03
pull request withdrawal symptoms

Vaern:
just minor changes to radiolysis shit, since pheo was playng us both

by the way, would it be possible to remove hazard traits (or at least not display them) in the case of values equal to or rounded down to 0? Lithium zirnox rods and some watz pellets suffer from it, but i'd rather not fuck with the hazard system just in case i break something


7pheonix:
3E0CB117-A2EE-4266-801A-E08EF5A9CB84

PR #592, opened 2022-04-29
Schrab fuel & waste + HEN recycling, gas centrifuge gating rebalancing

Vaern:
turns out hes was missing a "+5" and was registered twice


src/main/java/com/hbm/main/CraftingManager.java
```diff
@@ -757,7 +757,7 @@ public static void reg2() {
...
 		addRecipeAuto(new ItemStack(ModItems.upgrade_crystallizer, 1), new Object[] { "PHP", "CUC", "DTD", 'P', new ItemStack(ModItems.fluid_barrel_full, 1, Fluids.ACID.getID()), 'H', ModItems.circuit_targeting_tier4, 'C', ModBlocks.barrel_steel, 'U', ModItems.upgrade_centrifuge, 'D', ModItems.motor, 'T', ModBlocks.machine_transformer });
 		addRecipeAuto(new ItemStack(ModItems.upgrade_screm, 1), new Object[] { "SUS", "SCS", "SUS", 'S', STEEL.plate(), 'U', ModItems.upgrade_template, 'C', ModItems.crystal_xen });
-		addRecipeAuto(new ItemStack(ModItems.upgrade_gc_speed, 1), new Object[] { "PTP", "CBC", "PTP", 'P', CU.plate(), 'T', ModItems.upgrade_template, 'C', ModItems.coil_gold, 'B', (GeneralConfig.enableLBSM && GeneralConfig.enableLBSMSimpleCrafting) ? NB.ingot() : ANY_BISMOID.nugget() });
+		addRecipeAuto(new ItemStack(ModItems.upgrade_gc_speed, 1), new Object[] {"CNC", "GUG", "CMC", 'C', CU.plate(), 'M', ModItems.motor, 'G', ModItems.coil_gold_torus, 'N', NB.ingot(), 'U', ModItems.upgrade_template});
```
HbmMods: hough

Vaern: hough

7pheonix: houghopho

Vaern: as for a real answer, gating isotopic separation behind bismuth was the wrong choice for the gas centrifuges. It basically renders them useless, considering the SILEX + FEL are faster, better, and significantly less complicated options that are available at that point as well; plus, it prevents a full return on fluorite which is somewhat annoying.
Gating them behind niobium (rubber, in retrospect, might've been a better choice since I don't know if you can get NB from rare earth ore) would actually provide some utility for them during the oil and early nuclear stage. U-235's main applications at that point would be for fuel and for the Little Boy, which is fine as it's a low-power, primitive weapon that I've already made require more HEU to compensate. Besides, for something like my stillborn custom nuke rework ideas, it would add another layer of progression for pre-nuclear (shitty, low-power, low-tech U235 weapons > better ones using pu239, u233, et cetera) which I think would fit swimmingly.

HbmMods: niobium is post-shredder so it doesn't work as a gating material. the SILEX being infinitely better than the gas centrifuge can be fixed by simply changing the uranium separation requirement to UV (which i did), or we could make the centrifuge more attractive in general, like adding more upgrade slots or generally making them cheaper (although they are already pretty damn cheap)

Vaern: Honestly, I would still rather gate them behind rubber or maybe even technetium.

Vaern: > niobium is post-shredder so it doesn't work as a gating material. the SILEX being infinitely better than the gas centrifuge can be fixed by simply changing the uranium separation requirement to UV (which i did), or we could make the centrifuge more attractive in general, like adding more upgrade slots or generally making them cheaper (although they are already pretty damn cheap)

You bring up very valid points about making them less shit and such, but considering the fact that you can already get Pu-239 and U-233 from the ZIRNOX I'd argue that it would make the most sense to remove the bismuth gating still. I changed the copper plates out for rubber, that should work fine for gating; the latest I would gate it if that's a total impasse would be technetium steel, so at least it would be in line with the aforementioned reactor.
yuyuko-touhou

PR #605, opened 2022-05-14
use generalized spiral points algorithm

haru315:
Before

After

Ray is flown evenly regardless of size, and the previously existing blind spot on the top has been corrected.

### DIFF
--- a/src/main/java/com/hbm/explosion/ExplosionNukeRay.java
+++ b/src/main/java/com/hbm/explosion/ExplosionNukeRay.java
@@ -2,9 +2,6 @@ package com.hbm.explosion;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


MartinTheDragon (commented):
The code is a bit redundant and needs some cleaning, but this is definitely amazing!
It works very nicely with no extra perceived performance cost, and it may even be faster too.

HbmMods:
I just tested it on a redstone ready flat world with Little Boy, the block allocation time was a lot faster and the total time it took was marginally faster as well (2 minutes for the old code and 1:50 for the new one). The rays seem to retain their accuracy even up to the hard limit (especially noticeable with sandstone that has low blast resistance) with only a very small amount of floating blocks that remain. The new explosion code seems to be working really well!

PR #619, opened 2022-06-04
Modified RBMK fuels, added effects for boron/foam on RBMK and ZIRNOX

### LINKED ISSUE #620: Singleblock turbine recycling recipe
Used the latest HBM version.
I saw that anvil recipe that recycles small steam turbines gives 2 turbines (items), motors and tanks. But to craft the turbine you need 6 steel, 2 coils and turbine.
Seems that recipe that makes turbines was changed but the recycling didn't change.

Vaern:
CoC:

- Boron sand and foam effects on, along with changes to, destroyed reactors
  - BlockLayering can now be placed on the destroyed ZIRNOX
    - Using foam over the entire area will extinguish it much quicker than naturally, reducing the radiation output massively and preventing fire damage when close
    - Lowered chance of meltdown gas spawning
  - RBMK Debris
    - Lowered chance of meltdown gas spawning for both
    - Using boron sand on smoldering RBMK debris will help stop the fission reaction/cool down the debris much quicker than naturally
    - Using foam on burning RBMK debris will extinguish the flames much quicker than naturally
  - Reduced the BB size for certain debris types to apply contamination (no idea why it was *10*). Added RBMK graphite to the contaminating type, with the RBMK rod and ZIRNOX element types applying higher levels of contamination. Also, shredding for graphite.
- Dyatlov Meltdown Applicator can now be used on the ZIRNOX by shift-clicking
- Changed the localization for Negative-quadratic to dangerous, to more accurately reflect the level of caution necessary
- RBMK fuel stat changes
  - Neutron sources, LES now split into slow neutrons, due to the concentration of beryllium moderating fast neutrons
  - Pu238be and Po210be stats/functions changed respectively to make them actually useful
  - HEU233 and HEU235 changed to linear function/50 reactivity respectively, in order to respect their highly-enriched status
    - HEU235 is still square root in order to be consistent with HEN, which possesses a similar critical mass
  - MEN now splits with any neutrons; this is due to the build-up of Pu-239 from U-238 breeding, allowing both types of neutrons to jive with the Np-237/Pu-239 respectively
  - UEU heat changed to 0.65; I did not really like this change, since I wanted to keep the heat outputs dependent on enrichment, but for a low-power fuel like UEU, 0.5 is too much of a nerf. Either way, MEU will always have a higher heat output due to its higher flux function, so it works out.
- Fixed #620 by changing the recipe to match
- Fixed a little error in one of the NTM Ruins

### DIFF
--- a/src/main/java/com/hbm/blocks/machine/rbmk/RBMKDebrisBurning.java
+++ b/src/main/java/com/hbm/blocks/machine/rbmk/RBMKDebrisBurning.java
@@ -43,8 +43,9 @@ public class RBMKDebrisBurning extends RBMKDebris {
 				world.setBlock(x + dir.offsetX, y + dir.offsetY, z + dir.offsetZ, ModBlocks.gas_meltdown);
 			}
 			
-			//Foam helps stop the fire; 1.66% chance every 100-120 seconds for one side
-			int chance = block == ModBlocks.foam_layer || block == ModBlocks.block_foam ? 10 : 100;
+			//Foam helps stop the fire; Boron smothers it. 1.66% chance every 100-120 seconds for one side
+			int chance = block == ModBlocks.foam_layer || block == ModBlocks.block_foam ||
+					block == ModBlocks.sand_boron_layer || block == ModBlocks.sand_boron ? 10 : 100;
 			
 			if(rand.nextInt(chance) == 0) {
 				world.setBlock(x, y, z, ModBlocks.pribris);
--- a/src/main/java/com/hbm/blocks/machine/rbmk/RBMKDebrisBurning.java
+++ b/src/main/java/com/hbm/blocks/machine/rbmk/RBMKDebrisBurning.java
@@ -37,15 +37,14 @@ public class RBMKDebrisBurning extends RBMKDebris {
 			}
 			
 			ForgeDirection dir = ForgeDirection.getOrientation(rand.nextInt(6));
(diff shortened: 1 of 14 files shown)


src/main/java/com/hbm/blocks/machine/rbmk/RBMKDebrisBurning.java
```diff
@@ -36,12 +37,16 @@ public void updateTick(World world, int x, int y, int z, Random rand) {
...
+			Block block = world.getBlock(x + dir.offsetX, y + dir.offsetY, z + dir.offsetZ);
 
-			if(rand.nextInt(7) == 0 && world.getBlock(x + dir.offsetX, y + dir.offsetY, z + dir.offsetZ) == Blocks.air) {
+			if(rand.nextInt(10) == 0 && block == Blocks.air) {
 				world.setBlock(x + dir.offsetX, y + dir.offsetY, z + dir.offsetZ, ModBlocks.gas_meltdown);
 			}
 			
-			if(rand.nextInt(100) == 0) {
+			//Foam helps stop the fire; 1.66% chance every 100-120 seconds for one side
+			int chance = block == ModBlocks.foam_layer || block == ModBlocks.block_foam ? 10 : 100;
```
HbmMods: Shouldn't this include boron sand too?

Vaern: At the time, it made sense to restrict it just to foam, but in retrospect it doesn't really make much sense besides foam being cheaper I suppose. If necessary I can always change it later today.

PR #643, opened 2022-06-24
Progression bar issues with speed upgrades

Adam29Adam29:
Pheo reported this. It was a bug as the old progression check wasnt set to account speed upgrade changes


HbmMods:
Merged locally.

PR #680, opened 2022-07-15
Half-baked: The pull request.

UFFR:
Before making the PR official, I recommend testing and checking *everything*, many changes may have been done without my knowledge just to make the process faster, as I simply overwrote some of the files with ones from my fork so I didn't have to rewrite the same code 200 times, only manually editing if it caused errors. Some files may have been deleted not because I didn't want them, but because I couldn't be bothered to fix what was wrong with them. If you have any questions ask, I remember most intentional changes I have made. Also note, large chunks of code may be commented out. Those are features I thought too complex for a single PR, if you can believe that.
Most missing things are probably mundane things like lang entries and recipes, some more annoying ones like missing/wonky renders, but do check in case for more. Notify me of anything you find worthy of note.

List of changes in order of the directory structure:

- Addition: Computer file system classes.
    - Purpose for later use in computers.
    - Added to fix a dependency.
- Addition: Computer interfaces.
    - 3 levels of complexity in computer related machines.
    - IDevice is the simplest, for things like passive or automatic systems.
    - IComputer is the root for more complex machines, like actual computers.
    - ITerminal is a specification of IComputer, for terminal based commands. Not finished and under rework.
- Change: Static functions to IBatteryItem.
    - Give functions to create batteries filled or empty by default.
    - For use in recipes, so you can't cheese getting power.
- Addition: IDefenseAttributes interface.
    - API for working with NTM related defense or resistance properties.
    - Fairly self-explanatory.
- Addition: "Internet" related interfaces.
    - Not finished rebuilding.
    - For inter-computer communication.
    - Instead of transmitting a long, like the power net, they'll send either NBTCompoundTags or byte arrays that contain information of various kinds.
- Addition: Serialization registry.
    - Although the built-in Java serialization method exists, it is slow and bulky, though the interface does implement that too.
    - This is supposed to add a "universal" system to provide serialization to objects, usually for computer files.
    - Can also be used to store and recover other objects too.
    - Only a few of the classes use it, mostly to test it.
- Addition: Date class and IHasAge interface.
    - A wrapper for a BigInteger which can handle prolonged time periods.
    - Initial purpose was to have a "true" date held in the world, but as TEs or other entities went unloaded, a lot of time could pass and as if nothing had happend when the player returns. The system is intended to update in case of these time differentials.
    - For example, if you have an RTG running and leave for a long time, when you return it would update its internal pellets and deplete them accordingly without having to be constantly loaded.
    - Untested and unproven as of yet.
- Addition and changes: Removal of static integer GUI IDs.
    - Replaced with an enumerate, ID is the ordinal.
    - New GUIs can be added without worrying about overlapping IDs, since they are only used during runtime and are never needed to be stored.
    - The enum class is sort of like a registry. Has a standard behavior and 2 non-standard behaviors covered.
    - Also has an interface to allow custom or niche behavior.
- Change: Ammo types use subitems.
    - Reduces internal clutter and is more organized.
    - The old ones are removed completely, keep that in mind.
- Change: Turret blocks unified.
    - Since the only thing each turret class had different from the other was the provided TE and GUI ID, I just put that in the constructor.
- Addition: Nickel.
    - This was discussed for a bit. Cobalt-60 decays to Nickel, but Nickel is actually a useful element, so its addition wouldn't be too disruptive.
    - Most recipes aren't added yet or got yeeted because of bad merges, but it can quite literally fit in just about anywhere.
    - It's not super rare as an ore, but you can also get it from the asteroid miner.
- Addition: Activated Thorium (internally "THA").
    - An intermediary step in between plain Thorium-232 and Thorium fuel.
    - Highly radioactive and needs to be decayed, albeit for not too long.
    - Can be put in the radiation engine for extra power.
- Addition: Ferrouranium.
    - Finally added. Alloy between Uranium-238 and steel.
    - Used for high durability tools and armor, which also acts as a good hazmat.
    - Very dense and hard.
- Addition: Staballoy.
    - Alloy between Uranium-238 and Titanium.
    - Similar to ferrouranium, but lighter and more geared towards projectile weapons.
    - Also very hard but also chemically resistant, useful for high-grade DU penetrators.
- Addition: Orichalcum.
    - CBT incarnate. Optional, however.
    - Used for high-tier Lunarian weaponry.
- Unfinished: Bishamonten and Tsukuyomi.
    - These "turrets" caused a bit of a stir when I first tested them and showed them off in the server.
    - Bishamonten is a death ray, plain and simple. Has 3 modes, free target, where the target is manually given. Entity targeting, where it auto-tracks entities. And flood mode, where it gives out a burst of light that gives DRX to those that gaze into the abyss. Tends to kill everything loaded nearby. Unfinished, no model and the former two modes incomplete.
    - Tsukuyomi is like the Time Warp Rifle, but it doesn't care about the location of its target. Since the balance was all over the place, I thought about needing to have the target "tagged", in addition to the cooldown and expense. Works, technically, but not everything is finalized.
- Change: EasyLocation is more fleshed out.
    - The EasyLocation class was pretty bare bones when I first encountered it, so I remedied that. It is a simple implementation of the ILocationProvider interface.
    - Comparable/sortable, comparing which is closer to coordinates [0, 0, 0].
    - I added it to some pre-existing methods to prove its utility as an ease of use class.
    - It is serializable.
- Addition: EasyLocationStep.
    - Extended from EasyLocation with added methods to be iterable given a unit vector.
    - Unused and experimental, since bad things can happen if it cannot conclude properly (infinite loop).
- Addition: PreciseLocation.
    - EasyLocation but backed by BigDecimal, utility only hypothetical.
    - It is also serializable, but it uses the default Java serialization method so it may be slow.
- Change/Addition: Added config to GeneralConfig about rounded values.
    - Radiation values can be overly complex. This config makes it so it rounds it to so many significant figures.
    - Does not apply to the RBMK rods since more precision may be preferred.
- Change/Addition: 5 new potion types.
    - Paralysis: Intended to completely inhibit movement. Was harder to implement than I thought, so test in case of unexpected behavior. Used by the stun grenade.
    - Fragile: Increases the amount of damage taken. Given when suffering from radiation poisoning and some heavy bullets give it.
    - Unconscious (or subconscious): Koishi moment. Perfect invisibility from mobs. Though machines and some special mobs may not be affected.
    - Perforated: Heavy continuous damage effect, given by heavy bullets, since you're literally perforated by them. Can be nasty when paired with fragile.
    - Hollow: "Contaminated" but for Digamma.
    - Am particularly proud of the icons, except for paralysis, can hardly recognize it.
- Change/Addition: Added config to RadiationConfig about "realistic" RAD/s.
    - Calculated given the half-life in years divided by a constant.
    - Radioisotopes with a half-life less than 1 year have a boost in strength because they use a smaller constant.
    - Not balanced yet. As a note, it makes Tennessine-294 the most radioactive, *by far*.
- Change: Power armor is crafted without charge.
    - Seemed strange that they're first crafted with a full charge.
- Addition: Radiotherapy medication.
    - Adds more use to the rarer radioisotopes, but are in themselves unused.
- Addition: Ferrouranium tools and armor.
    - High durability, used as the base in upgrade based paths.
- Change: Renamed IRON and GOLD DictFrame objects.
    - Now FE and AU respectively, to be consistent with the others.
- Change: Nuclear creeper uses safe nuke if mobGriefing is false.
    - Looks like they were supposed to, but it doesn't work, so I fixed it, hopefully.
- Addition: Bullet armor penetration.
    - In its infancy stage, more of a proof of concept.
    - Bullets with a penetration value that fall below the armor protection level do not do any whatsoever. If it just barely fails to penetrate (within 20%), it deals substantially less damage. However, if it manages to penetrate, it deals full damage.
    - Penetration drops off gradually with distance by default. Custom behavior over range can be specified with a functional interface using a lambda.
- Addition: EntityMappings class.
    - Functionally the same as the TileMappings class.
    - Works for both mobs and technical entities.
    - Registry is very simple with redundancies removed when possible.
- Addition: Lunatic explosion type.
    - As mentioned with Bishamonten, it gives Digamma to those within its range, effectiveness varies wildly with its strength.
- Change: Simplification and tweaking of gun factories.
    - The manufacturer field is an enumeration now, to reduce redundancies and to add localization, if applicable.
    - An optional "equip" sound effect. Currently, only the M2 Browning uses it so far.
    - The gun's name field is also a field for localization.
    - Bullet configs retrieve from immutable lists, for further redundancy removal and simplification.
    - There is a string field for "damage", if it is not empty, the line specifying the gun's damage will use that instead.
    - 2 lists for "lore" and "function". Basically Wikipedia stuff. Hidden unless pressing [shift] or [ctrl], respectively.
- Addition: Benelli M4 autoshotgun.
    - A classic weapon. 8 round tube magazine, full auto capability. High durability.
    - Has a variant with a 24 round drum magazine, though it has no distinguishing model.
- Addition: .45 ACP caliber.
    - Bit stronger and more expensive than 9mm, but similar usage in handguns and SMGs.
    - Used in the Thompson and UAC pistol.
    - More planned uses.
- Addition: UAC Sidearm.
    - The .45 pistol from PB3. Not too fancy.
    - Has iron sights.
- Addition: G36 rifle.
    - Not too much to it, simple 5.56mm rifle.
    - Low durability, but useful if you want a proper AR fast.
- Addition: Medium Lunatic Rifle.
    - Another 5.56mm AR, part of the Lunar Defense Corp lineup.
    - High durability and decently high capacity magazine of 45 rounds.
    - Has iron sights.
- Addition: Lunatic Marksman Rifle.
    - Very powerful anti-material rifle, part of the Lunar Defense Corp line of weapons.
    - Default round is a 20mm DU sabot that should kill everything you're likely to encounter instantly. Very fast and accurate projectile + very high durability.
    - Originally single shot, but changed to have a small 4 round internal magazine.
- Addition: M2 Browning HMG.
    - The classic, needs no introduction.
    - Fast firing, very high durability, belt-fed, and a blast to use.
    - Uses the Howard reload sound as its equip sound effect.
    - Can be held in a fashion as if it had iron sights.
- Addition: .308 Winchester / 7.62x51mm NATO caliber.
    - More powerful version of 5.56mm, for use in heavier rifles.
    - Planned uses in the UAC DMR, UAC Carbine, UAC LMG, and M60 machine gun.
- Addition: Light Lunatic Rifle.
    - Another Lunar Defense Corp weapon.
    - Fast firing, high capacity, 9mm SMG.
    - Has iron sights.
- Change: 240mm shells are more powerful.
    - The APFSDS ones also have very high penetration values. If you're using them against tanks or ships I guess?
- Addition: True/full scale 240mm nuclear round.
    - Self-explanatory.
    - We did too much trolling.
- Addition: Time Warp Rifle.
    - One of the main features of my fork in the early days, finally works.
    - Magazine capacity of 16 micro-singularities, semi-auto function.
    - Hitscan weapon, impossible to dodge if properly aimed and fired.
    - Damage should bypass armor and directly damage, should be instant kill for most players.
    - Model is a bit rough, but it's accurate to the source.
- Addition: Lunatic gun and scattergun.
    - Unfinished, the former is a sidearm and the latter a shotgun. Both energy weapons.
- Change: Most bullet behaviors are now lambdas.
    - Experimental, should still work, but change back if they don't.
- Change: Unused Combine weapon configs.
    - Mostly for testing, initially planned for NPCs, but never got that far.
    - Remove at will.
- Deletion: Deleted GunPoweredFactory.
    - GunEnergyFactory already exists + way too many errors for me to go through.
    - Restore/fix as necessary.
- Addition: Blast Furnace IMC.
    - Since you were crazy about that a bit back I guess, so I added one for the blast furnace since I could.
- Addition: Chemical Plant IMC.
    - It's for my chemical plant recipe handler, so it probably won't work for yours right away.
- Change: NEI Blast Furnace handler uses my recipe handler now.
    - Only made sense.
- Addition: EnumGUI enumerate class.
    - Explained in an above note.
    - Basically acts as a registry.
    - Since the only the ordinal of the enumeration is used at runtime, there is no worry of ID collisions.
    - Allows for custom behavior.
    - Some GUIs may not work, probably from bad merges. Shouldn't be too hard to fix though.
- Change: FuelHandler uses a Map.
    - Simple change, works just fine.
- Change: GUIHander updated to use EnumGUI.
    - Self-explanatory.
    - Has error handling and logging in case of errors, which should be rare.
    - Much neater now.
- Change: GunConfiguration class tweaks
    - Better JavaDoc on some fields.
    - Some fields have default assignments to prevent accidental NullPointerExceptions.
    - Unload ability, untested and probably doesn't work, just means that you can remove bullets from the gun.
    - Is now cloneable, if anyone needs that.
- Addition: GunConfigurationEnergy class.
    - Functionally the same as the original class, but acts just as you'd expect.
    - Charge time, ammo power consumption rate, and max charge editable.
- Change: HazardTransformerRadiationNBT has static constructor method.
    - Makes working with this feature *a lot* easier.
    - Currently used for only a few recipes.
- Change: HazardTypeBase's addHazardInformation method specifies the type of its List<> parameter.
    - It annoyed me and the IDE, still works fine.
    - (Now if only I could tell the IDE to ignore it in overrides...)
- Change: HazardTypeBase subclass tweaks.
    - Hot and radioactive dropped items have a field of effect around them.
    - I expect it to be somewhat laggy if there's a lot of them. Optimize if possible.
    - Nifty to harm entities this way.
- Change: HazardRegistry has more values.
    - For the heavy actinides mostly.
- Change: HazardSystem tweaks.
    - The Maps have their types specified.
    - New method for updating dropped items.
    - 4 radiation type proportion values stored in a single float value?!?! Satisfactory!
    - Most hazard entries don't use it yet, so they default to being beta emitters.
- Deletion: ICustomWarhead interface.
    - Some mistakes are best forgotten.
- Addition: IFastChargeable interface.
    - Like suit batteries, but more generic and potentially reusable.
    - Initial purpose was to charge energy guns, but might be used for anything I suppose.
    - I should add more types that have predicates for better specification in what they charge or not, but it's not use anymore, so that can go on hold.
- Addition: IHasLore interface.
    - ItemCustomLore implements this, use the interface itself for when that class can't be used.
- Addition: ILocationProvider interface.
    - Self-explanatory, interface that holds coordinates.
    - Also has utility methods and some static constructors for ease of use.
    - EasyLocation is a simple implementation of this interface, use that for most cases, but use the interface as storage just in case.
- Addition: Explosive fluid type.
    - Explodes (proportional to the fill) when broken with 1,000mb or more inside. Untested.
    - Only nitroglycerin uses it so far.
- Unfinished: Sarin gas fluid.
    - First chemical weapon. Unused, but does have the new "chemical" fluid hazard icon.
    - Planned to use another fluid type for when it breaks in a container, but never got to that.
- Unfinished: Molten salt fluid.
    - For the test MSR, but that never got done, so it's unused for now.
- Unfinished: Chlorine triflouride fluid.
    - Doubles as a nasty weapon and industrial chemical!
- Unfinished: Hydrobromic acid fluid.
    - Meant for the acidizer rework to be paired with Pheo's electrolyzer, but never developed that far, so again, unused.
    - Could also probably used as a chemical weapon.
- Unfinished: Sulfuric acid fluid.
    - Redundant now, you should remove it assuming it's not used anywhere else.
- Unfinished: Warp plasma fluid.
    - For the warp reactor and eventual FTL.
    - Unused because I didn't include the warp reactor, which is *very* broken at the moment.
- Unfinished: Nitroglycerin fluid.
    - Use in recipes and funny exploding barrels.
- Unfinished: Nitric acid fluid.
    - Used to make nitroglycerin.
- Change: Upgraded the check for incongruent fluid registries.
    - Now tells the name of the fluid that's missing from one of the registries.
    - Only added because I found the need for this feature when testing the merge, but it's useful I guess.
- Addition: Constructor methods for fluid tanks.
    - Minor thing used for recipes.
- Change: FluidType class tweaks.
    - The internal collection for the traits is a now a Set<> to prevent duplicates.
    - Setting the temperature of a fluid autoadds the appropriate temperature related trait (hot or cryo).
    - Added method to automatically get the localization.
    - Trait list for the tooltip is autogenerated, sorta. It's a little funky but it's no longer an if statement chain.
    - The trait enumeration has the radioactive, biohazard, and chemical traits.
- Change: GUIs for RTG related machines use the getter for the pellet list.
    - Protection for the list.
- Addition: Various anvil recipes.
    - Check for ones that I accidentally removed!
- Unfinished: New battery recipes.
    - Still pending for balance, but was so they were both more realistic and less crap from what we have now.
- Change: Various assembler recipes.
    - Again, check for ones that I accidentally removed!
- Unfinished: Assembler recipe class type.
    - Unused, but is mainly used to test the researchable class/interface, more on that later.
- Addition: Blast furnace recipe handler.
    - As I promised to do.
    - Supports items, blocks, item stacks, and ore dictionary.
    - Also has IMC support.
- Addition: Chemical plant recipe handler.
    - Well, I made it before you did, but you added yours, so yeah.
    - I think it was required by something, but I don't remember.
    - It has unique recipes near the end.
- Change: Tweaks to OreDictManager.
    - Concrete now has ore dict, mostly for simpler recipes.
    - Iron and Gold use their chemical symbol for internal consistency.
    - Nether quartz may also be known as regular quartz.
    - Protactinium-233, though I don't know any other mod that adds Pa-233, mostly for hazard registry.
    - Uranium-234, though I don't think it still exists in the fork.
    - Activated Thorium, mostly for hazard registry.
    - Heavy actinides, more on that later.
    - "NbBe", Niobium-Beryllium alloy, one of those Niobium things Pheo wanted.
    - "NbSuper", Niobium superalloy, the big Niobium thing Pheo and the others wanted.
    - Ore dict for the Australium "isotopes".
    - Records have ore dict, because for some reason that's a thing.
- Change: DictFrame supports subitems and radiation types.
    - It has an internal HashMap<> for registring the hazard multipliers based on the prefix.
    - The Optional<> is for the multi-item.
    - The enum vararg method is to register the ore dict entries based on the supplied enums, thus the prefix registry.
    - The enum must also implement an interface to override this behavior, if applicable.
    - Look for the examples.
- Addition: The AStack implementations are serializable.
    - ComparableStack, ComparableNBTStack, and OreDictStack.
- Change: AStack now has 2 new abstract methods.
    - getFriendlyName() provides a human readable name, usually the localized name, but so long as it is easily identified.
    - toStack() just makes it an ItemStack.
    - The NBT related ones you can delete, since they got superseded by the serialization.
- Change: ItemBattery now extends ItemCustomLore.
    - Minor thing, but now it supports lang tooltips if applicable.
- Addition: ItemBatteryFast class.
    - Used by the IFastChargeable interface.
    - Generic battery item for quickly charging inventory items like tools and energy guns.
    - Supports single-use and reuseable types.
    - Code is a bit clunky, so in need of a rework.
- Change: ItemRTGPellet implements IRadioisotopeFuel.
    - So it is compatible with the IRTGUser interface.
- Change: ItemRTGPellet pellet registry privated.
    - For protection purposes.
    - It is accessible through a getter that puts it into an ImmutableList<>, doesn't change use cases.
- Change: ItemRTGPelletDepleted has a Curium type.
    - For the Berkelium RTG pellet.
- Addition: ItemStorageMedium class.
    - Unused but is planned for the computers.
    - Has several possible types, using subitems to reduce item bloat and simplifying logic.
    - Stores its files in NBT as byte arrays.
    - Has hashes of the files too in case of corruption. Also is neat.
- Change: ItemWasteLong's Schrabidium type renamed.
    - Don't remember why, but it now specifies the isotope.
- Change: Ditto for ItemWasteShort.
- Removal: GunMP class.
    - T'was unused.
- Removal: GunNightmare class.
    - Ditto.
- Change: ItemAmmo uses subitems.
    - *Significantly* reduces bloat and simplifies logic.
    - No more internal if statement chain either, autogenerated.
- Change: ItemGunBase updated.
    - Updated to conform with the updates to the GunConfiguration class.
    - Also specifies the bullet damage and penetration of the current config in the tooltip.
- Addition: ItemGunEnergyBase class.
    - Basically the former but for energy guns.
- Addition: ItemAmmoEnums enumeration collection class.
    - Has all of the enum types for ammo, standardized.
    - Also has the grenades.
- Addition/Change: Hand grenades are subitems too.
    - Logic is stored in a registry and only the enum is passed around.
    - Most logic isn't registered yet, unregistered grenade types default to the stock one.
- Change: ItemCustomLore updated.
    - Mostly for utility methods and autogeneration.
- Change: ItemEnumMulti tweaks.
    - Added types to the enums, minor thing.
- Addition: ItemEnumMultiSimple class.
    - The ItemEnumMulti but the subitem unlocalized names use the enum name and ignore the main item's unlocalized name.
    - Mostly for lazy reuse of lang and textures.
- Addition: IItemEnum interface.
    - Overrides for the aforementioned classes, optional.
    - Adds overrides for the tooltip, ore dict, shimmer effect, rarity, hazards, and texture.
- Unfinished: Item enum types.
    - Most probably unused, but keep for later use.
    - Orichalcum, has a lot of subitems, so this removes the bloat.
    - Raw storage, raw forms of data storage mediums, like magnetic tape and hard drive platters.
    - Ammo assemblies.
    - Scintillators, for a future project.
    - CPUs, unused items, for the computers.
    - RAM, ditto.
    - Einsteinium-254.
    - Californium.
    - Berkelium.
    - Curium.
    - Nickel, unused.
    - Protactinium.
    - Semiconductor wafers.
    - Silicon.
- Unfinished: ItemStorageEnums enumeration collection class.
    - Same as the former but specifically for the storage types.
- Addition: Various minor items.
    - Silicon, circuit related.
    - Ac227Be: like Ra226Be, but much stronger.
    - Sa327Be: gives use for Solinium.
    - Cf252Be: extremely strong neutron source.
    - Forged Dineutronium, intermediary for an Orichalcum type.
    - Rare earth and Tennessine catalysts, ditto, but could be used elsewhere too.
    - Micro-singularity, TWR ammo type.
    - Activated Thorium Zirnox rod.
- Change: Schrabidium fuel renamed to MES.
    - For internal consistency.
- Addition: HbmCollection class.
    - Library for commonly used collections and other values.
    - Meant to be generic and free of logic, to keep such things out of specific classes.
    - A bit bloated, but it does its job.
- Change: Library class tweaks.
    - Player UUIDs are marked final.
    - Contributor Set<> marked final.
    - getLifespan() moved here.
    - radiate() function, for all your irradiation purposes.
    - easyPrintHook(), was added so I could add those print hooks for like the cracker, but you already did that on your end.
    - boolToByte() and byteToBool(), converts a boolean array to a byte array and vice versa with the latter having a default value.
    - compressStringCollection() and decompressStringBytes(), serializes a Collection<String> to a byte array and vice versa.
    - autoNBT(), takes any object and tries to find the appropriate NBT type for it.
    - roundDecimal(), uses a BigDecimal to round a decimal value to so many sigfigs.
    - roundNumber(), ditto, but made by Drillgon200 and should work for any number without the need of a new BigDecimal object.
    - getBlink(), checking for the blinking for certain things.
    - convertScale(), converts a number, given the min - max, to another number, given a new min - max.
    - toPercentage(), converts a proportion to a string percentage.
    - doesArrayContain(), concatArrays(), randomFromArray(), randomFromCollection(), and randomPrime(), self-explanatory.
    - easyTargetPoint(), generates a TargetPoint given a TileEntity or ILocationProvider object.
    - filledArray(), generates an array of the given size of a given type. Exists only because Arrays.fill() didn't work for something I don't remember.
    - arrayToList(), self-explanatory and probably not needed.
    - EnumHashAlgorithm, all supported hash algorithms in Java.
    - getHash(), generates a hash given the bytes.
    - doHashesMatch(), just MessageDigest.isEqual().
    - toHexString(), converts a byte array to a hex string.
    - iterableToString(), turns any iterable object to a single string, in the format of: [(object), (object), ...].
    - A isObstructed() overload that uses ILocationProvider.
- Change: New ModDamageSource types.
    - Bleed: Damage source for the Perforated potion. Bypasses armor.
    - TWR: Mostly deprecated, for the TWR of course. Bypasses armor.
    - Light: Digamma type of damage. Bypasses armor and is allowed in creative mode.
- Change: Clay catalysts use Nickel instead of Iron.
- Addition: DeserializationException class.
    - Exception wrapper for errors occuring during deserialization.
- Addition: KoishiBreachException class.
    - :koishicup:
    - Comes with a random message and popup.
- Change: MainRegistry check.
    - Attempts to find items and blocks with a null registry name.
    - Not sure it works well.
- Unfinished: New NEI handlers.
    - Arc furnace, research reactor, and decay drum.
    - Forgot to register them, don't know if they work.
- Change: EnumSymbol additions.
    - Chemical: Chemical hazard, specifically chemical weapons.
    - Biohazard: Looks surprisingly decent at the resolution.
    - Hot: Hot tamale.
- Change: renderAmmo(), renderItemAndEffectIntoGUI(), and renderAmmoAlt() tweaks.
    - Parameter for the item changed to ItemStack.
- Addition: TimeSavedData class.
    - Meant to track the world's time based on my system.
    - Unused since I removed it from the world tick event handler, because I thought it was causing lag.
    - Can be readded I suppose, but it's unused at the moment.
- Change: RTG machine TEs implement IRTGUser.
    - The whole point of the interface.
- Change: Base NT turret TE uses AStack for its ammo.
    - To compliment the change to the GunConfiguration class.
    - Should work since it's otherwise the same class besides that one change.
- Change: IRadioisotopeFuel updated.
- Change: IRTGUser updated.
- Change: BobMathUtil additions.
    - Static final fields for data storage magnitudes.
    - durabilityBarDisplay(), self-explanatory.
    - remainingRadioactiveMaterial(), experimental function based around an equation used to calculate the amount of radioactive material remaining, given the original amount, the half-life, and the time that has passed.
    - radiologicalDating(), similar to the above, but to date objects given the remaining proportion of material and its half-life.
    - entityUnitVector(), calculates the unit vector from an entity, repurposed from the bullet code.
    - offsetForHand(), modifies an EasyLocation object to be offset to the hand, also repurposed from the bullet code.
- Addition: HashWrapper<> class.
    - Experimental class designed to give built-in classes equals() and hashCode() functionality.
    - Methods are provided as lambdas or method references.
- Change: InventoryUtil additions.
    - doesPlayerHaveAStack(), ease of use function.
    - countAStackMatches(), ditto, but specifically made for the reworking of ItemGunBase.
- Change: LoggingUtil color types now an enum.
    - Works the same, but now more like the chat formatting enum.
- Addition: NBTBuilder class.
    - Wrapper for NBTCompoundTag object.
    - Methods return itself for easier chaining.
- Addition: NBTSerialization class.
    - Has methods for serializing all the NBT objects.
    - Untested, big, and scary.
- Addition: ObjectWrapper<> class.
    - Wrapper but for all main built-in Java object methods.
- Deletion: RTGUtil class.
    - What IRTGUser is for.
- Addition: SerializationWrapper<> class.
    - Another wrapper but for serialization.
- Addition: StringWrapper<> class.
    - Wrapper to add the toString() method.
- Addition: TimeDurationType enumeration.
    - Extracted from ICustomWarhead.
- Addition: Wrapper<> interface.
    - Interface for all wrapper types.
    - Forces override of equals(), hashCode(), and toString().
    - Has abstract method getStoredObj().
- Change: US lang file.
    - Check for missing entries!
- Addition: New models.
    - Tesseract: The counter-resonant singularity said something about tesseracts you shouldn't look at, so I planned on actually adding them.
    - Atomic clock: Not yet added.
    - Watz Supercompressor: The one and only.
    - Third Eye aka Satori: Not yet added.
    - Weapon models including a spear.
- Addition: New sounds.
    - Some button sounds.
    - The alarm/alert sound effect from DEFCON.
    - Numerous Combine chatter sounds. For when I get to that.
    - Casing bounce sounds from PB3.
    - Nuclear Mission Jam: For the associated record.
    - Silencio from OFF: Not added, but is intended to occasionally play in the distance when the player had DRX exposure.
    - Keyboard typing sound effects: For the key inputs in GUI.
    - Various weapon sound effects from PB3.
- Addition: Some unused GUI textures for features I didn't add yet because they were too complicated and would make this PR too big.
    - 4 Warp Core GUIs.
    - The Bishamonten and Tsukuyomi GUIs.


ShimmeringBlaze:
Good god my guy, that's almost an entire mod in itself

UFFR:
> Good god my guy, that's almost an entire mod in itself

Don't worry, there's still more to come.

src/main/java/com/hbm/items/ModItems.java
```diff
@@ -7151,6 +7150,8 @@ private static void registerItem() {
 		GameRegistry.registerItem(ammo_44, ammo_44.getUnlocalizedName());
 		GameRegistry.registerItem(ammo_5mm, ammo_5mm.getUnlocalizedName());
 		GameRegistry.registerItem(ammo_9mm, ammo_9mm.getUnlocalizedName());
+		GameRegistry.registerItem(ammo_45, ammo_45.getUnlocalizedName());
+		GameRegistry.registerItem(ammo_308, ammo_308.getUnlocalizedName());
```
70000hp, line 7154: You forgot to register the 357 magnum

PR #696, opened 2022-07-28
Config comments, satellite registry changes

Toshayo:
Small change in satellite registry, added config entries comments


Toshayo:
Changing access modifier of the satellite register function is necessary to allow creating custom satellites without reflection.

PR #752, opened 2022-09-24
Nuke improvements

70000hp:
COC

Nukes:

The general blast power of nukes has been raised (not radius) ,so they'll finally be able to pierce concrete, reinforced stone, the lot

The effect that a nuke's radius has on its blast power is actually noticeable now, weaker nukes (little boys, fat mans,) will not be able to break ducrete , while themonukes can.

Fallout:

After a blast, Concrete near the center has a chance to become damaged, getting a lower blast res and becoming affected by gravity, Same effect happening with concrete bricks and ducrete (at a lowered chance and it does not fall down when cracked)

Dead leaves can now be set on fire because why not?

Blocks:

Added Concrete and Ducrete Debris 

Small changes to a few recipes of blast resistant blocks

Broken Concrete bricks now fall down 

Balefire:

Balefire now emits environmental radiation because it didn't make sense before according to horse book

CMB tiles and above can withstand a balefire blast so now they have an actual use


70000hp:
Sure love it when lang decides to screw up the merge for no reason, fixed

PR #766, opened 2022-10-12
Nuclear boiler

PAP6:
Adds the ability to heat oil and cracked oil in the RBMK heating column and cool them in the RBMK heat exchanger

### DIFF
--- a/src/main/java/com/hbm/tileentity/machine/rbmk/TileEntityHeatex.java
+++ b/src/main/java/com/hbm/tileentity/machine/rbmk/TileEntityHeatex.java
@@ -74,6 +74,8 @@ public class TileEntityHeatex extends TileEntity implements IFluidAcceptor, IFlu
 	public static FluidType getConversion(FluidType type) {
 		if(type == Fluids.MUG_HOT)		return Fluids.MUG;
 		if(type == Fluids.COOLANT_HOT)	return Fluids.COOLANT;
+		if(type == Fluids.OIL)		return Fluids.HOTOIL;
+		if(type == Fluids.CRACKOIL)	return Fluids.HOTCRACKOIL;
 		return Fluids.NONE;
 	}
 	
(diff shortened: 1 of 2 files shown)


HbmMods:
Just a few problems:
* The `heatCap` value isn't set for any of the oil types which defaults them to 0.01, while coolant has 0.25, which means the same amount of heat will boil 25x more oil than it does coolant. That's good if you want to process a ton of oil, but terrible for cooling a nuclear reactor.
* `TileEntityHeatex`' conversion is backwards, the hot input is defined as oil and the cold output as hot oil. Best case scenario it means that hot oil just won't work as a heat exchanger input, worse case is that inserting cold oil reverses the heat exchanger's process and turns steam back into water.
* Oil in general isn't suitable for cooling a reactor since if the oil is actually used in a refinery, the reactor ends up with open cycle cooling that fully relies on pumpjacks to keep it filled, which isn't a great idea. If the oil is just used as a heat medium in a closed cycle, then it seems redundant since regular coolant already does the same thing.

Recommended course of action:
There's no real use in fixing the implementation as it is now. The current RBMK heat exchanger was always just a placeholder anyway, once the electric heater is released I'll start working on a larger heat exchanger that uses Thermal Units, which combined with the TU boiler I made last week would allow you to heat oil and also power other heat-accepting machines.

PAP6:
I'm sorry
1. I'm sorry for not knowing about the default heatCap
2. I'm sorry for being dumb and blind
3. The main idea is to let the oil become hot faster so I could process it faster since electric boilers are not fast enough for me, not to cool the reactor. But it still could be useful in some reactors since oil's boiling temperature is just 350 celsius, while the mug root beer and coolant's are much higher. I'm sorry

Sorry for the inconvenience and wasting your time, I'll stop trying to turn my shitty ideas into reality

HbmMods:
Regarding the boilers, the ones we have now will become obsolete since the way they work conflicts with the new standard. The new heat-powered boiler is much larger and can be powered by any heat source, which means that by using either a higher tier fuel or by stacking up electric heaters it'll be possible to boil oil much faster.

PR #787, opened 2022-10-30
Consolidated the Geiger Counter into the singular block

Vaern:
CoC:
- Removed the geiger_counter item and replaced all occurences with its block equivalent
- Created the ItemBlock ItemGeigerCounter instead
  - Maintains removed item's functionality (overlay, right clicking)
  - Places block when sneaking


Vaern:
pash :(

PR #856, opened 2022-12-24
Removed fibrosis, pirfenidone, gloves; reverted reacher functionality

Vaern:
CoC:
- Removed rubber gloves, pirfenidone
  - Reacher reverted to use pre-glove functionality, except for the radiation function
  - Made ItemPill use i18n for descriptions
- Removed pulmonary fibrosis
  - The effects of radon and others actually work with mesothelioma as is, so they now increment that instead


Vaern:
unknown-3

PR #858, opened 2022-12-27
Gradle improvements

Toshayo:
The first commit fixes the project not loading when there is no cache on local system. It has been tested on a fresh-installed VM with InteliJ IDEA 2022.3.1. The maven repository has been taken from Mekanism 1.7.10 Community Edition (caus i had issues with NEI downloads).

The second commit improves the version specification. Instead of reading from mcmod.info it takes the version and build number from gradle properties. If the build number is specified (not empty), the output version is, for example, 1.0.27_X4445. If the build number is empty, the version will be 1.0.27


HbmMods:
The original build script had a couple of lines (27-52) that would automatically add all NEI-related dependencies as referenced libraries in eclipse as part of the setup process. Does that still happen with the new version of the script or will dependencies have to be added as referenced libraries manually?

Toshayo:
I've just tested with Eclipse 2020-06 (4.16.0). It has successfully downloaded NEI-related stuff and added to classpath ("implementation" function)
image

Toshayo:
PR https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/pull/897 does the same and better.

Issue #895, opened 2023-01-26
[Workspace] Workspace creation issues 


Shibva:
Following the contribution directions I downloaded the src files from the forge page and merged the files from NTM to the forge source files; I then tried to use BASH to setup the decompiled workspace but when running I ran into this error

```
Execution failed for task ':downloadClient'.
> java.io.IOException: Server returned HTTP response code: 403 for URL: http://s3.amazonaws.com/Minecraft.Download/versions/1.7.10/1.7.10.jar
```

This could just be me but I might need a second opinion

Shibva:
if I recall correctly, forge did a fuckup and the original files are no longer avalable at that locatioin. if this is the case then the instructions would need to be updated.

martemen:
Hi @Shibva,
I had the same problem and fixed it on my [fork](https://github.com/martemen/Hbm-s-Nuclear-Tech-GIT), the first two commits fixed the problem. [Commit one](https://github.com/martemen/Hbm-s-Nuclear-Tech-GIT/commit/1ecd569a28e299751d1d36e0962dfe077e39b401), [Commit two](https://github.com/martemen/Hbm-s-Nuclear-Tech-GIT/commit/77a029f8b4835176cd2ef5443bbd5b72ad1db1cc). I plan to merge my repo back to this one at some point.

Shibva:
Does it come with IDE intogrstionn or just the fix?

martemen:
It works for me in Intellij idea.

PR #897, opened 2023-01-28
Issue #895: Gradle fix

martemen:
A fix for the ForgeGradle compilation error from #895.

**Changes:**

- Updated the Gradle versions from 3.0 to 4.4.1
- Changed the ForgeGradle from the official to [anatawa12's fork](https://github.com/anatawa12/ForgeGradle-1.2)
- Changed the encoding to Unicode


Toshayo:
Isn't https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/pull/858 doing the same? :)

Toshayo:
My fork does not change encoding to Unicode though

martemen:
My fix works with Gradle 4.4.1. 7.4.2 has many [deprecations](https://docs.gradle.org/current/javadoc/deprecated-list.html).

HbmMods:
I know next to nothing about gradle so I'm not sure what PR to merge, what even is the functional difference between gradle versions? And has this been tested with Eclipse?

martemen:
@HbmMods It wasn't tested with Eclipse but I do expect this to work.
@Toshayo Which pull request you prefer? Do you prefer to combine my changes with yours in PR #858?

Toshayo:
I think @martemen gradle fix is better because:
 - it does not use a random maven found on the net
 - the gradle version has less deprecations

However, I think exporting all variables like version and stuff may be handy. Actually, I think it is to @HbmMods to decide if they prefer the old way of specifying version or the gradle.properties one :).

For now I'll test the fix with Eclipse, though I'm pretty sure there will be no problems with

martemen:
> I think @martemen gradle fix is better because:
> 
> * it does not use a random maven found on the net
> * the gradle version has less deprecations
> 
> However, I think exporting all variables like version and stuff may be handy. Actually, I think it is to @HbmMods to decide if they prefer the old way of specifying version or the gradle.properties one :).
> 
> For now I'll test the fix with Eclipse, though I'm pretty sure there will be no problems with

**Thank you!**

Toshayo:
Everything works as expected.
 - The project loads
 - External dependencies are present
 - Project builds
 - Runs are runnable
Used Eclipse 2020-06 (4.16.0)

martemen:
@Toshayo I added Gradle properties, although I removed certain parameters like Minecraft versions, or mod versions because most of those won't probably be updated (the[ lasted forge 1.7.10 release](https://files.minecraftforge.net/net/minecraftforge/forge/index_1.7.10.html) is 7 years old).

Toshayo (approved):
I'm closing my PR

PR #913, opened 2023-02-13
mixer custom recipe(test)

sdddddf80:
Most of the custom recipe codes are come from LiquefactionRecipes. Now we can add or change mixer's recipe by editing hbmMixer.json. But somehow the "writerecipe" part doesn't work, which means we have to write all the recipe in the hbmMixer.json by ourselves, or there won't be any recipes in mixer. How does "writerecipe" part work @HbmMods ? Now I have no ideas. Can you give me a hand?


sdddddf80:
Here is a workable hbmMixer.json
[hbmMixer.txt](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/files/10718450/hbmMixer.txt)

HbmMods:
`writeRecipe` will most likely fail for most if not all recipes because usually either one of the fluid inputs or the solid input is `null` which won't work in the JSON serialization methods.

sdddddf80:
That means if `rec.getValue().solidInput` is null,then `writer.name("inputItems");this.writeAStack((ComparableStack) rec.getValue().solidInput,writer);` will fail. Is that right?

HbmMods:
Correct, both `writeAStack` and `writeFluidStack` don't accept `null` values, they will throw an exception causing the config writer to stop.

sdddddf80:
New issue. Now I can write all the recipe into hbmMixer.json, but it can't read any recipe which doesn't have inputItem or one of the inputfluids, even if I have add the comparative statement.Here is the java file.
[MixerRecipes.txt](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/files/10723567/MixerRecipes.txt)

HbmMods:
Use `has("name")` of your recipe's `JsonObject` to check if an input is available or not.

sdddddf80:
Like'JsonObject obj = (JsonObject) recipe;if(obj.has("inputItems")){}`?

sdddddf80:
Like `JsonObject obj = (JsonObject) recipe;if(obj.has("inputItems")){}`?

sdddddf80:
`this.readAStack(obj.get("inputItems").getAsJsonArray())`will also fail if `obj.get("inputItems").getAsJsonArray()` is null,right?

sdddddf80:
Finally, it is done.

HbmMods:
Merged locally, although I took the liberty of simplifying the read/write methods a fair bit.

For example, checking and then casting to either `ComparableStack` or `OreDictStack` for the `setSolid` method is redundant since it accepts `AStack` which is the superclass for both of them.

And while I was at it, I also added configs for oil fractioning and catalytic cracking as well, since all those handlers share similar recipe structures.

PR #942, opened 2023-03-08
more custom recipe support 

sdddddf80:
Add custom recipe support for breeder,press,crystallizer and combination. Enable lens to see bedrock ore.


HbmMods:
Had to merge locally because coincidentally earlier today I also did the press recipe configs.

PR #951, opened 2023-03-12
fluid pump

Voxelstice:
this pull request basically adds a fluid pump block
- it can pump water if there's a water block below it, pumps less water if its a flowing block
- same for lava
- it cant pump any other fluid
- you can right click on the fluid pump with a fluid identifer to change what it should pump

the texture is bad as i cant draw properly, i am also not sure if the recipe is balanced enough

i also wanted to get this tooltip thing working but i didnt get it working despite putting the right code


SoB1nary:
why

src/main/java/com/hbm/blocks/ModBlocks.java
```diff
@@ -3200,6 +3204,7 @@ private static void registerBlock() {
 		GameRegistry.registerBlock(machine_forcefield, machine_forcefield.getUnlocalizedName());
 		GameRegistry.registerBlock(radiorec, radiorec.getUnlocalizedName());
 		GameRegistry.registerBlock(radiobox, radiobox.getUnlocalizedName());
+		GameRegistry.registerBlock(fluid_pump, fluid_pump.getUnlocalizedName());
```
HbmMods: You should use `register(fluid_pump)` here, not only is it shorter but it also automatically sets the ItemBlock to `ItemBlockBase.class` which is necessary for the tooltip to work.

src/main/java/com/hbm/blocks/machine/MachineFluidPump.java
```diff
@@ -0,0 +1,115 @@
...
+
+public class MachineFluidPump extends BlockContainer implements ILookOverlay, ITooltipProvider {
+
+	public MachineFluidPump(Material mat) {
+		super(mat);
+	}
+
+	@Override
+	public TileEntity createNewTileEntity(World world, int meta) {
+		meta = 0;
```
HbmMods: Meta is set to 0 but never used?

src/main/java/com/hbm/blocks/machine/MachineFluidPump.java
```diff
@@ -0,0 +1,115 @@
...
+		
+		List<String> text = new ArrayList();
+		text.add((pump.power < pump.getMaxPower() / 20 ? EnumChatFormatting.RED : EnumChatFormatting.GREEN) + "Power: " + BobMathUtil.getShortNumber(pump.power) + "HE");
+		text.add(EnumChatFormatting.RED + "<- " + EnumChatFormatting.RESET + I18nUtil.resolveKey("hbmfluid." + pump.tanks[0].getTankType().getName().toLowerCase()) + ": " + pump.tanks[0].getFill() + "/" + pump.tanks[0].getMaxFill() + "mB");
+		
+		ILookOverlay.printGeneric(event, I18nUtil.resolveKey(getUnlocalizedName() + ".name"), 0xffff00, 0x404000, text);
+	}
+	
+	@Override
+	public void addInformation(ItemStack stack, EntityPlayer player, List list, boolean ext) {
```
HbmMods, line 111: I don't think this part is going to work without the ItemBlock being set correctly

src/main/java/com/hbm/tileentity/machine/TileEntityFluidPump.java
```diff
@@ -0,0 +1,223 @@
...
+
+import api.hbm.energy.IEnergyUser;
+import api.hbm.fluid.IFluidStandardTransceiver;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraftforge.common.util.ForgeDirection;
+import net.minecraft.block.Block;
+import net.minecraft.init.Blocks;
+import net.minecraft.world.World;
+
+public class TileEntityFluidPump extends TileEntityMachineBase implements IFluidAcceptor, IFluidSource, IEnergyUser, IFluidStandardTransceiver {
```
HbmMods: `IFluidSource` is part of the deprecated old fluid system and shouldn't be implemented in all newer TEs.

`IFluidStandardTransceiver` is for TEs that both have in and output tanks, but the pump only outputs so it should be `IFluidStandardSender` instead.

`TileEntityMachineBase` is for TEs with slots, you should extend `TileEntityLoadedBase` and implement `INBTPacketReceiver` for packet sync.

src/main/java/com/hbm/tileentity/machine/TileEntityFluidPump.java
```diff
@@ -0,0 +1,223 @@
...
+	
+	public long power = 0;
+	public FluidTank[] tanks;
```
HbmMods: Tanks don't need to be an array when there's only one.

src/main/java/com/hbm/tileentity/machine/TileEntityFluidPump.java
```diff
@@ -0,0 +1,223 @@
...
+			this.updateConnections();
+			setupTanks();
+			
+			if(hasPower() && canPump() && tanks[0].getMaxFill() > tanks[0].getFill()) {
+				int amount = getPumpAmount();
+				int convert = Math.min(amount, tanks[0].getMaxFill() - tanks[0].getFill());
+				
+				if (convert > 0) {
+					tanks[0].setFill(tanks[0].getFill() + convert);
+					power -= this.getMaxPower() / 20;
```
HbmMods: That's equivalent to 50 assemblers. Heavy infinite tanks are half as good and don't need any power at all.

src/main/java/com/hbm/tileentity/machine/TileEntityFluidPump.java
```diff
@@ -0,0 +1,223 @@
...
+			if(b == Blocks.flowing_lava)	return 250;
+		}
+		
+		return 0;
+	}
+	
+	public void pump() {
+		Block b = worldObj.getBlock(xCoord, yCoord-1, zCoord);
+		
+		if(b == Blocks.lava || b == Blocks.flowing_lava || b == Blocks.water || b == Blocks.flowing_water) {
```
HbmMods: Check seems redundant since `pump` already requires `canPump` to return `true`.

HbmMods:
Overall I'm not entirely sure as to what the pump is intended to be, water can already be obtained easily through infinite tanks and lava is effectively useless. Since the pump only detects blocks directly below it, it's not for draining large areas either. The only use I could imagine would be 528 mode while disabling infinite tanks, although the energy consumption would still be way too high for earlygame.

Voxelstice:
the fluid pump was made just to add a little bit of 'realism' to the mod, it is useless on it's own but stacking it can lead to some high water outputs which i think is useful for some people

i could make it drain larger areas, but im still not exactly sure how to do this without making the code a mess

PR #982, opened 2023-04-01
Small fixes

TehTemmie:
1. Oil wells now actually benefit from using the overdrive upgrade (The old behavior meant that overdrive only affected the speed decrease from power-savings upgrades, and using overdrive + power-savings was still slower than not using either upgrade to begin with while also consuming as much, if not more, power)
2. RBMK fuel items that exceed their max temperature no longer cause reactor temperatures to skyrocket exponentially when DisableMeltdowns is enabled. (Old behavior meant that RBMK rods would have their temperature increased by the average heat instead of being made equal to the average heat of the three components)


HbmMods:
flopthumbsup

PR #1004, opened 2023-04-19
More uses for the heat system

sdddddf80:
Add heat gain mechanism for liquefactor,vacuum distill,refinery and fraction tower. Now these machines can get heat to obtain different gains. Players can also now use electric heater to control the heat of the machine.They have to determine the heat at the highest machine efficiency by observing the displayed data.(ps:I am not sure that whether the gain function is balanced and players can not use the electric heater to finely control the heat of the machine.)

### DIFF
--- a/src/main/java/com/hbm/blocks/machine/HeaterElectric.java
+++ b/src/main/java/com/hbm/blocks/machine/HeaterElectric.java
@@ -75,17 +75,25 @@ public class HeaterElectric extends BlockDummyable implements ILookOverlay, IToo
 		TileEntityHeaterElectric heater = (TileEntityHeaterElectric) te;
 
 		List<String> text = new ArrayList();
-		text.add(String.format("%,d", heater.heatEnergy) + " TU");
-		text.add(EnumChatFormatting.GREEN + "-> " + EnumChatFormatting.RESET + heater.getConsumption() + " HE/t");
-		text.add(EnumChatFormatting.RED + "<- " + EnumChatFormatting.RESET + heater.getHeatGen() + " TU/t");
-		
+        if(!heater.mode)
+		{
+			text.add(String.format("%,d", heater.heatEnergy) + " TU");
+			text.add(EnumChatFormatting.GREEN + "-> " + EnumChatFormatting.RESET + heater.getConsumption() + " HE/t");
+			text.add(EnumChatFormatting.RED + "<- " + EnumChatFormatting.RESET + heater.getHeatGen() + " TU/t");
+		}
+		else {
+			text.add( I18nUtil.resolveKey("info.hold_heat"));
+			text.add(String.format("%,d", heater.heatEnergy) + " TU");
+			text.add(I18nUtil.resolveKey("info.hold") + String.format("%,d", heater.holdheat) + " TU");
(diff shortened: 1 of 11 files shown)


70000hp:
Greg.

70000hp:
Jokes aside, it doesn't seem to have a whole lot of purpose outside of adding an extra thing a player has to worry about when making refineries
And given new reactors the heat system is already plenty used

HbmMods:
Wdym "worry about", I thought these additions were optional

sdddddf80:
> Jokes aside, it doesn't seem to have a whole lot of purpose outside of adding an extra thing a player has to worry about when making refineries
> And given new reactors the heat system is already plenty used

Well,I just want players to have more opportunities to combine machines for use in more than just generating electricity.

HbmMods:
Just a few more questions:

* What exactly is the electric heater's new mode and why is `holdHeat` a separate value?
* Why do the new heat consumers all have special hardcoded cases for the electric heater? Wouldn't it make more sense to return `holdHeat` through the standard `getHeatStored` function if the mode is enabled? After all it does add quite a bit of boilerplate that is identical over most of them.
* Why do some of the machines have a hard-cap on the heat they can store, while others don't? Wouldn't it make more sense to have a percental loss over time, while certain values below a soft cap no longer provide any benefit?
* Both the liquefactor and the vacuum refinery (but strangely not the regular refinery) have hard caps on the heat they can buffer, wouldn't it make more sense to show the heat in a GUI gauge instead of a tooltip? Spreading out information over both a GUI and a tooltip isn't that good, the only machine that does it is the heat exchanging heater and that's because the heat is uncapped, so having a gauge wouldn't even be possible.

sdddddf80:
> Just a few more questions:
> 
> * What exactly is the electric heater's new mode and why is `holdHeat` a separate value?
> * Why do the new heat consumers all have special hardcoded cases for the electric heater? Wouldn't it make more sense to return `holdHeat` through the standard `getHeatStored` function if the mode is enabled? After all it does add quite a bit of boilerplate that is identical over most of them.
> * Why do some of the machines have a hard-cap on the heat they can store, while others don't? Wouldn't it make more sense to have a percental loss over time, while certain values below a soft cap no longer provide any benefit?
> * Both the liquefactor and the vacuum refinery (but strangely not the regular refinery) have hard caps on the heat they can buffer, wouldn't it make more sense to show the heat in a GUI gauge instead of a tooltip? Spreading out information over both a GUI and a tooltip isn't that good, the only machine that does it is the heat exchanging heater and that's because the heat is uncapped, so having a gauge wouldn't even be possible.

1.this mode will let the electric heater hold the heat of the machine which is above it as the value `holdHeat` .
2.em, I can try that.
3.I remember all the machines have a 12000TU heat Hard-cap, and in hold heat mode, the max heat that the electric heater can hold is 1000TU so there is no need to set a heat hard-cap.
4.I just think that is more audio-visual for players to find the law.

sdddddf80:
Well it looks like that new heat consumers have to have special hardcoded cases for the electric heater as in "Holdheat" mode, the value 'heat' will be assigned as value ‘holdHeat’, not add value ‘holdHeat’.

PR #1013, opened 2023-04-29
Fluids editor

sdddddf80:
add hbmFluids.json which allows players to edit the fluids' attribute, including the fluid it can be heated to and the fluid it can be cooled to .It also support the reload command.


JameH2:
me omw to add the "corrosive" trait to water

sdddddf80:
What specific effect do you want to achieve?@HbmMods

HbmMods:
Closing as it's redundant with the most recent commit

PR #1041, opened 2023-05-26
Burningwater202

burningwater202:
I have no idea what I'm doing


src/main/java/com/example/examplemod/ExampleMod.java
HbmMods, line 1: please don't include this

src/main/resources/assets/hbm/lang/en_US.lang
HbmMods, line 1: encoding machine broke

HbmMods:
I'm just going to delete the offending files, I hope the delete option works the way I think it does because github's documentation is piss poor and most gitoids aren't exactly helpful either.

HbmMods:
nope it doesn't do that at all

burningwater202:
GitHub my behated

HbmMods:
ow my legs

PR #1045, opened 2023-05-27
Basement blocks

sdddddf80:
Add 3 basement blocks. One can just play an alarm and summon 10 FBIs when the player is within 30 squares, which can summon 5 times, and will only summon again when the number of nearby FBIs is less than 2. Among the remaining two, in addition to summoning the FBI,  when the player is between 80 and 30 squares, one will launch Incendiary missile and the other will launch cluster missile.If player break or explode the basement block, it will summon 20 FBIs at once. But if player right-click it with a  screwdriver, it will harmless turn into air and drop 10 enhanced circuit. Now only the one which only summon FBIs will generate in Radio structure as I am not sure which structure is suitable for the remaining two. Also add config options for whether FBI can break blocks.


ncr-t:
nnqpcqyr8bx71

70000hp:
Guh.

PR #1050, opened 2023-06-03
Sorry for being a pane

burningwater202:
this is my Glass't pull request I swear

(for now)

-Added BlockNTMGlassPane for handling glass panes

-Added Reinforced Laminate Glass Panes - Crafting recipe is the same as normal glass panes, may need to change if balance is a worry (or just modify the block resistance so that they're less blast resistant as panes)

-Added Reinforced Glass Panes because why not

Laminate Glass Panes are currently clear- which does make sense a little bit since thinner glass would allow more light through and potentially be clearer (I'm no scientist but I'm gonna just pretend I know what I'm talking about)
This is mostly because I have the dain bramage and can't get the opacity to work on the panes. Oh well

Probably going to mess with the code for connecting the pane blocks to other blocks in the future as right now it will only not connect to air

Also thinking of adding a glowing glass variant that utilizes Tritium to have a slight (~lv8 light level or so) glow to them, maybe maybe not


JameH2:
For every pull request you accept, a nuclear missile is launched towards a random country, pull requests like these help starving nuclear missiles grow and flourish in heavily populated countries, 

Pull request now and help a missile find its home and maybe even a billion lives to happily call its own <3.

burningwater202:
Just send it directly to my address instead

.gitignore
```diff
@@ -26,3 +26,9 @@ run
 
 # Changelog backup
 /changelog.bak
+CREDITS-fml.txt
+forge-1.7.10-10.13.4.1614-1.7.10-changelog.txt
+LICENSE-fml.txt
+MinecraftForge-Credits.txt
+MinecraftForge-License.txt
+README.txt
```
Toshayo: Why these files?

HbmMods:
image

PR #1065, opened 2023-06-09
Bunkers (for real)

Vaern:
CoC:
 - Bunkers !!
   - Procedurally-generated, with varying amounts of rooms between bunkers
   - Uses existing MapGenNTMFeatures out of convenience
   - Bit labyrinthine, but that can't be helped yet
   - Replacement for the old bunkers
 - ProceduralStructureStart
   - A replacement for the bloated and anti-OOP ProceduralComponents
   - Much easier to use and actually overridable (compared to the static mess taht was the aforementioned class)
   - Comes with some more documentation, so it's not trial and error for offsets
 - Reworked NTMWorldGenerator
   - Discovers if the relevant biome has a population event (like how superflats do not)
     - If they do, it creates structures before vanilla population and post-generation, preventing conflicts with flora and other natural features
 - Some structure meta methods (doors actually work now)


HbmMods:
image

PR #1070, opened 2023-06-12
Minor addition in pollution code. (clamping values)

BallOfEnergy1:
Saw a comment in the discord about it, thought I might as well fix it myself because I knew the issue.
Clamps the soot and heavy metal pollution to a minimum value. 

(also some updates to OC integration)


HbmMods:
If this is about the display thing, wouldn't it make more sense to employ rounding on the pollution sensor than to change pollution itself?

BallOfEnergy1:
It might make more sense to round here, let me see if I can try and do that.

Edit: I'm not too familiar with java still, so I'm having trouble finding a way to round/truncate to a certain decimal point easily.

PR #1071, opened 2023-06-12
Curseforge publishing improvements

Toshayo:
This PR improves metadata sent while publishing the mod on CurseForge. Display name is customized and differs from the filename, dependencies to Forge, Java 8 are added. The mod is also marked as client and server side.

The CurseForge configuration sample is also brought back as it is nicer to have a commented sample file than trying to get required fields from code.

There's also a fix for the changelog backup file not being replaced, generating an exception.

Also a smol improvement in potion code replacing decimal numbers to hexadecimal for more readable colors.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,8 @@
+import org.gradle.plugins.ide.eclipse.model.internal.FileReferenceFactory
+
 import java.nio.file.Files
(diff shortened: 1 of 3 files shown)


HbmMods:
very nice

PR #1077, opened 2023-06-18
Enhance RBMK Tungsten Carbide Neutron Reflector

sdddddf80:
Enhance RBMK Tungsten Carbide Neutron Reflector. Now it can be set to reflect neutrons in the specified direction.This will allow players to redirect excess neutrons and utilize them, such as concentrating them for irradiation.


funnyman8:
This seems a little unbalanced for how cheap the reflectors are, perhaps it could be a new type made with graphite blocks and bismuth ingots? That'd at least put it in line with some of the other higher-tier RBMK recipes.

Vaern:
Whilst the code for this PR is already dodgy (a setting *and* a mode, as well as the redundant switch code), the premise just doesn't mesh very well with RBMK.

Troodontid:
It also doesn't make sense from a physics viewpoint. IRL neutron reflectors don't reflect neutrons like light bouncing bouncing off a mirror. Instead, they scatter neutrons in every direction. Also, this seems like a very exploitable thing, as you can simply focus the output of a bunch of RaBe rods onto a single irradiation column to instantly complete recipes.

BallOfEnergy1:
As stated by others, this seems very cheaty and could be exploited thoroughly using multiple rods feeding into one column and doesn't really feel like a good idea in this case. Also, the addition doesn't fit into the realistic premise, as neutron reflectors don't work as a mirror, as stated by 238-Pu. I don't think it'd be a good addition, as it would break realism along with adding a way to break progression in unintended ways.

BallOfEnergy1:
Also, the code looks very odd, as there's 2 different mode arrays, mode and mode1, along with that there's a setting, which makes virtually no sense here.

sdddddf80:
add a new block as the Enhance RBMK Neutron Reflector.But somehow the Cover Panel installed on it won't display

PR #1084, opened 2023-06-23
Wiring errors with slightly more info

Voxelstice:
it's a very small change that just adds more info on why the pylons didn't connect

i don't know if im doing this right, it worked just fine in testing

### DIFF
--- a/src/main/java/com/hbm/items/tool/ItemWiring.java
+++ b/src/main/java/com/hbm/items/tool/ItemWiring.java
@@ -68,7 +68,7 @@ public class ItemWiring extends Item {
 							player.addChatMessage(new ChatComponentText("Wire end"));
 
 						} else {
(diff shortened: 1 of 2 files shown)


HbmMods:
1. Wouldn't it make more sense to change `canConnect` to return an integer instead and using the int code to generate a message instead of using a string buffer?
2. Not a fan of the all lowercase messages

Voxelstice:
> Wouldn't it make more sense to change ``canConnect`` to return an integer instead and using the int code to generate a message instead of using a string buffer?

never considered doing that, i usually do it the way i did in this pr, i'll rewrite the code slightly to be done like this

> Not a fan of the all lowercase messages

i don't really know how you want it to be, you can always edit the messages later, but for now i simply made the first letters of those messages uppercase

PR #1117, opened 2023-07-17
Added OC compatibility to yet another block (more coming soon)

BallOfEnergy1:
Added OC compat to the fluid gauge to measure fluid flows.


HbmMods:
Doesn't most OC code run serverside? I don't think calling I18nUtil there is a good idea

BallOfEnergy1:
Ah, didn't think about that, I'll change it back to the way it was before.

BallOfEnergy1:
there jesus fucking christ my IDE wants to murder me; finally got it working
also added comparator support for tanks/barrels because 70000hp kept bugging me about it

70000hp:
> there jesus fucking christ my IDE wants to murder me; finally got it working also added comparator support for tanks/barrels because 70000hp kept bugging me about it

:3

PR #1138, opened 2023-08-11
structural gain for custom machine

sdddddf80:
Add structural gain for custom machine. Now by using higher-level structural blocks allowed in the configuration file, the efficiency of machine operation will be improved.


sdddddf80:
Added restrictions. Now only the custom_machine series blocks will offer the gain. By the way, the reason why I made this is that I don't like Thermal Expansion, Mekanism style machine upgrade philosophy: Want a better machine? Dismantle the original and build a new one! I want to encourage players to personally upgrade a machine to its best state step by step, rather than simply categorizing the machine into beginner and advanced categories

PR #1151, opened 2023-08-16
Fixes for Crucible (Thermos fork)

re7gog:


re7gog:
They deleted thermos.Thermos class in fork so we can use another

PR #1159, opened 2023-08-23
Resistance Rework !!

Vaern:
CoC:
- Squashed the vast majority of resistances down into the hundreds
  - Concrete, Concrete Bricks, Reinforced Stone etc. can now be damaged by conventional explosives
  - Resistance is based on tiering and microcrafting complexity
    - For instance, Reinforced Stone is immune to all but the strongest conventional explosives (Blocks of C4, Semtex, HIMARS, etc.). Compound Mesh blocks all (current) conventional explosive strengths due to its gating behind solidification.
  - Comes with little-no disruption of nuclear defense
    - The function behind nukes means that most blocks' survivability are regardless of their resistance being in the hundreds or thousands


70000hp:
you forgot one 
image
other than that, scrumptious!

Vaern:
can't do anything about that i'm afraid, vanilla obsidian is 2000 * 3/5 = 1200

PR #1164, opened 2023-08-26
Make conveyor crane I/O sides configurable

MartinTheDragon:
Screenshot_2023-08-26_15-36-54

Right click with a screwdriver to change output direction, shift-right-click to change input direction.

The tons of textures were made by 70k. Many thanks to him!

PR #1165, opened 2023-08-26
Improve crane I/O screwdriver logic

MartinTheDragon:
Instead of cycling through all directions, the side the player clicks on will be manipulated directly, and, if applicable, input will be swapped with output and vice versa.

Right-clicking normally changes the conveyor side, shift-right-clicking the machine side.

Clicking on a side twice will manipulate the opposite side.

Related to #1164.


HbmMods:
flopthumbsup

PR #1166, opened 2023-08-27
Obsidian balancing

70000hp:
Nukes can now crack obsidian into obsidian gravel, making it significantly less overpowered 
They can also crack obsidian bricks into cracked obsidian bricks, which have much lower blast resistance, and will turn into obsidian gravel upon a hit
Obisidian bricks are now slightly more expensive, requiring tungsten bolts instead of iron bars


JameH2:
Smh bruh why not use the catalytic cracker

HbmMods:
why don't we just make obsidian weaker to nuclear explosions to be consistent with all the other blast resistant blocks?

70000hp:
Not sure how would I change the blast resistance of a vanilla block, if I knew that possible I wouldn't have gone with this funky workaround

70000hp:
The issue is that normal obsidian would be stronger than the bricks

MartinTheDragon:
> why don't we just make obsidian weaker to nuclear explosions to be consistent with all the other blast resistant blocks?

I don't like that solution. You're changing vanilla behavior that other mods might depend on. At least make a config option to turn that off, if you really don't want the cracking mechanic. I'd prefer it not only because of that, but also because it may make weapons that create multiple, smaller nuclear explosions more interesting, and in turn, also make NTM's high resistance blocks actually considerable. Because currently, obsidian just completely overshadows any of those.
Three birds with one stone, actually.
(Also I wouldn't be able to port this because block properties are immutable in modern versions lol)

HbmMods:
I'm not saying we should change obsidian's blast resistance value, I'm saying that we should change obsidian to be weaker to *nuclear explosions*, i.e. use a lower resistance value when obsidian intersects with a ray. I did the same thing to sandstone actually where instead of being almost as weak as air, I made the explosion interpret it as strong as regular stone.

MartinTheDragon:
Why not both then? I like the cracking mechanic even in that case, because it won't just instantly atomize your obsidian if you expected it to hold. Maybe even make an API interface for blocks that crack into other blocks, so this mechanic can be used by other things and explosives as well.

HbmMods:
Personally not a fan of the cracking mechanic, and people generally don't build bunkers out of obsidian, at least none that I have ever seen.

70000hp:
The cracking mechanic in this case is designed to be a big drawback, in my other usages of it like concrete I added caulk as a fluid, which allows for easy repairs via chemthrower, I could probably add caulk as a means to repair the damaged blocks

70000hp:
> I'm not saying we should change obsidian's blast resistance value, I'm saying that we should change obsidian to be weaker to _nuclear explosions_, i.e. use a lower resistance value when obsidian intersects with a ray. I did the same thing to sandstone actually where instead of being almost as weak as air, I made the explosion interpret it as strong as regular stone.

Seems fair enough, Ill make another PR with the intercept if you've made up your mind on the cracking

HbmMods:
I already added block-specific overrides yesterday with sandstone, adding obsidian was just one extra line.

70000hp:
Issue is solved I presume, you just need to nerf obsidian bricks now

PR #1170, opened 2023-08-28
Faster conveyors

BallOfEnergy1:
yeah just make the higher conveyor tiers move faster, the texture files were changed due to the fact that using the `.mcmeta` file made it skip frames and just in general look crap.
(haven't tested it on a server, though it should operate mostly the same)

there 70000hp, you happy?


70000hp:
:3333333333333
very, thankie

70000hp:
Even though faster conveyors don't increase throughput, they reduce entity count, which is fairly important since large setups tend to lag, and just feel better to use

BallOfEnergy1:
(and it also gives a reason for multi-lane conveyors to exist)

HbmMods:
grafik

PR #1175, opened 2023-08-29
Ammo variant tweaks

70000hp:
**Changes:**

Adjusted all ammo variant crafting recipes to match their stock ammo crafting amounts, e.x if you've crafted a round which is made in stacks of 32, the variant will be made in stacks of 32
- This acts as a buff to several guns, making their ammo variants cheaper to craft

Chlorophyte ammo now penetrates targets, causing it to jump between mobs if their target dies
- It also has had it's recipe yield doubled, since it was not worthwhile to craft before
- While it is not as faithful to the source material, it is a significant buff to an ammo type which was nigh useless


HbmMods:
Did you actually test the chlorophyte bullet penetration? Because the target isn't reset on impact and the targeting angle is now 360°, I don't see how the bullet would seek another target instead of just bouncing back and forth around the initial target.

HbmMods:
Also wouldn't a targeting angle of 360° mean that the bullet would come out of your gun backwards if there's an entity behind you which is just a little closer to you than what you're actually trying to shoot?

70000hp:
> Did you actually test the chlorophyte bullet penetration? Because the target isn't reset on impact and the targeting angle is now 360°, I don't see how the bullet would seek another target instead of just bouncing back and forth around the initial target.

Oh shoot, I thought I reverted that

70000hp:
I'll reset the target when the entity impacts, the bullet bouncing around its initial target helped it score multiple hits on the same entity, which I thought was fairly decent damage-wise

70000hp:
And yes I tested the penetration extensively, it was very good at clearing out large waves of mobs, while not leaving any of them half-damaged due to the bullet jumping about the target

70000hp:
https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/assets/105080577/016ee8b5-8101-4793-8ab3-7c2ab6879221

Here you go, they now dont hit stuff behind them and are much better at killing swarms

70000hp:
@HbmMods just incase you haven't seen, I fixed the chloro rounds issue

PR #1177, opened 2023-08-31
Animations for Josh

70000hp:
**Changes:**

Added:

- Animations for both reloading and recoil for the josh
- Rescaled the model to be consistent with the bolter in terms of scale and orientation, so animating wouldn't be annoying, which is why git says the model changed.
- It is a very good way to make animating easier, I would recommend it whenever you implement a new gun.

Fixed:

- The UV issues in the mag.


70000hp:
Here's a vid showcasing the anims, the burst fire function is not included with this PR

https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/assets/105080577/0da92e7e-3887-4d55-b988-bf409f1644ff

HbmMods:
Rescaling the OBJ doesn't actually achieve a whole lot because the actually difficult part of animating is getting the timing and magnitudes right, the scale is just a static multiplier for the movement which can be changed on the fly. Scaling in the code is also much easier than rescaling the entire model, especially when there's multiple moving parts and the movement axis have to be precise.

BallOfEnergy1:
:letsfuckinggoooooooo:

PR #1181, opened 2023-09-01
Burst fire

70000hp:
**Added**

Burst-fire firing mode to the gun system

- Self-explanatory, Change the firingDuration parameter to adjust how long does each individual shot takes, and roundsPerBurst to adjust how much shots per action
- Currently added as an alt-fire to the Josh (pretty much selective fire, prevents you from wasting 6 shots on one glyphid scout) and the UAC pistol (since it had that option in project brutality)

### DIFF
--- a/src/main/java/com/hbm/handler/GunConfiguration.java
+++ b/src/main/java/com/hbm/handler/GunConfiguration.java
@@ -25,6 +25,8 @@ public class GunConfiguration implements Cloneable {
 	public int rateOfFire;
 	//amount of bullets fired per delay passed
 	public int roundsPerCycle;
+	/** Amount of rounds per burst, irrelevant if not a burst fire weapon**/
+	public int roundsPerBurst;
 	//0 = normal, 1 = release, 2 = both
 	public int gunMode;
 	//0 = manual, 1 = automatic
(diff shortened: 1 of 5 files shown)


HbmMods:
`burstDuration` can't be a simple field because that would mean every gun of the same type shares this value because items of the same type share instances. Either use NBT or one of the already existing timers.

70000hp:
> `burstDuration` can't be a simple field because that would mean every gun of the same type shares this value because items of the same type share instances. Either use NBT or one of the already existing timers.

Oh right, completly forgot, I'll add it to NBT

70000hp:
> `burstDuration` can't be a simple field because that would mean every gun of the same type shares this value because items of the same type share instances. Either use NBT or one of the already existing timers.

@HbmMods There we go, I converted it to use NBT

HbmMods:
The UAC pistol has a burst size of 3, yet fires 6 times

70000hp:
Six? I've used it and it only fired 3 shots per burst

70000hp:
https://quickshare.samsungcloud.com/a0pAydJjZlP9

HbmMods:
https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/assets/16674258/bdbc6abb-fa15-4048-95c2-daebb5dd0d76

70000hp:
I just tested it in main ntm and it fired 3 shots like it was intended to
what

https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/assets/105080577/53f1ffa4-efea-4dfb-9382-4d6fa1c9277f

PR #1182, opened 2023-09-03
Cargo arty shell fix

70000hp:
Self-explanatory, I had to fix this for a server so I might as well PR it


70000hp:
Fixes a crash caused by the cargo artillery shell, caused by not calling super.entityInit in EntityArtilleryShell

HbmMods:
Already fixed that earlier today in-dev

PR #1183, opened 2023-09-03
Glyphids part 1, Infested buildings

70000hp:
**Added**

Infested buildings:
- Glyphid hives with an open design, nested within buildings
- Within them, you can find loot in the form of scrap metal and gun parts
- In lore, glyphids would be acting somewhat like ants and bringing things back to their hive, which you could then loot
- Amount of hives per building varies depending on the size of the building
- Their spawn rate is based on the amount of chunks with soot in the world, with the base chance being configurable
- Whether they should spawn at all is also configurable


70000hp:
Pictures of the hives
image
image
image
image
and the general loottable 
image

src/main/java/com/hbm/entity/mob/EntityGlyphidScout.java
```diff
@@ -100,7 +100,7 @@ public void onUpdate() {
 				if(b.getMaterial() != Material.air && b.isNormalCube() && b != ModBlocks.glyphid_base) {
 					this.setDead();
 					worldObj.newExplosion(this, posX, posY, posZ, 5F, false, false);
-					GlyphidHive.generate(worldObj, (int) Math.floor(posX), (int) Math.floor(posY), (int) Math.floor(posZ), rand);
+					GlyphidHive.generateBigGround(worldObj, (int) Math.floor(posX), (int) Math.floor(posY), (int) Math.floor(posZ), rand);
```
MartinTheDragon, line 103: The manual `Math.floor` operations are unnecessary. Casting floats/doubles to integer types automatically truncates them.

src/main/java/com/hbm/world/feature/GlyphidHive.java
```diff
@@ -123,6 +216,44 @@ public static void generate(World world, int x, int y, int z, Random rand) {
...
+
+					switch (block) {
+						case 1:
+							world.setBlock(x + i - 5, y + j - 2, z + k - 5, ModBlocks.glyphid_base);
+							break;
+
+						case 7:
+						case 8:
+							world.setBlock(x + i - 5, y + j - 2, z + k - 5, ModBlocks.deco_loot);
+							if (block == 8) LootGenerator.lootMakeshiftGun(world, x + i - 5, y + j - 2, z + k - 5);
```
MartinTheDragon, line 240: It might make more sense to have the `LootGenerator` functions themselves place the block instead to avoid this extra check, but I guess this is the internal pattern

src/main/java/com/hbm/world/feature/GlyphidHive.java
```diff
@@ -123,6 +216,44 @@ public static void generate(World world, int x, int y, int z, Random rand) {
...
+
+		for(int i = 0; i < 11; i++) {
+			for(int j = 0; j < 7; j++) {
+				for(int k = 0; k < 11; k++) {
+
+					int block = schematicBigStructure[6 - j][i][k];
+
+					switch (block) {
+						case 1:
+							world.setBlock(x + i - 5, y + j - 2, z + k - 5, ModBlocks.glyphid_base);
```
MartinTheDragon, line 234: These coordinate calculations are the same throughout the entire function, so why not extract it into a variable to make the code shorter?

src/main/java/com/hbm/util/LootGenerator.java
```diff
@@ -93,7 +93,37 @@ public static void lootCapStash(World world, int x, int y, int z) {
...
+
+						scrap = type < 2 ? ModItems.bolt_tungsten : type == 2 ? ModItems.wire_tungsten : ModItems.bolt_dura_steel;
+						loot.addItem(new ItemStack(scrap), world.rand.nextGaussian() * 0.06, i * 0.01125,  world.rand.nextGaussian() * 0.06);
+					}
+				break;
+
+				default:
+					int type2 = world.rand.nextInt(8);
+					scrap = type < 2 ? ModItems.hull_big_steel : type2 < 4 ? ModItems.hull_big_titanium : type2 < 7 ? ModItems.pipes_steel : ModItems.tank_steel;
+					loot.addItem(new ItemStack(scrap), world.rand.nextGaussian() * 0.06, 0, world.rand.nextGaussian() * 0.06);
```
MartinTheDragon, line 123: Are you sure you didn't miss the y-motion there?

src/main/java/com/hbm/world/gen/component/Component.java
```diff
@@ -663,7 +673,44 @@ protected void fillWithBlocks(World world, StructureBoundingBox box, int minX, i
 			}
 		}
 	}
-	
+	/** Have you ever wanted Glyphids in your building? No? Too bad! **/
+	protected void infest(World world, StructureBoundingBox box, int minX, int minY, int minZ, int maxX, int maxY, int maxZ, int maxSize, Random rand){
+		if(!enableInfestation)
+			return;
+
+		PollutionHandler.PollutionPerWorld ppw = PollutionHandler.perWorld.get(world);
```
MartinTheDragon, line 681: This could work, but makes world generation also dependent on when something was generated. There will be different results for the same seed (which is not that much of an issue), and it would punish players that play on low render distance and reward those who play on high or run around a lot.

src/main/java/com/hbm/world/gen/component/OfficeFeatures.java
```diff
@@ -532,7 +534,8 @@ public boolean addComponentParts(World world, Random rand, StructureBoundingBox
 			generateInvContents(world, box, rand, ModBlocks.filing_cabinet, getDecoModelMeta(0), 10, 5, 12, HbmChestContents.filingCabinet, 4);
 			generateInvContents(world, box, rand, ModBlocks.filing_cabinet, getDecoModelMeta(0), 10, 6, 12, HbmChestContents.filingCabinet, 4);
 			generateInvContents(world, box, rand, ModBlocks.filing_cabinet, getDecoModelMeta(0), 2, 9, 5, HbmChestContents.filingCabinet, 4);
-			
+			fillWithRandomizedBlocks(world, box, 4, 14, 0, 8, 14, 0, rand, ConcreteBricks);
```
MartinTheDragon, line 537: Are you sure this belongs here?

MartinTheDragon (changes requested):
Mostly clean-up suggestions, but also notices on some maybe-bugs (no pun intended).

src/main/java/com/hbm/util/LootGenerator.java
70000hp: It only spawns one so item in this case, so it doesn't matter

src/main/java/com/hbm/world/gen/component/Component.java
70000hp: Wdym?

70000hp: To be fair, the infested buildings are not overly difficult, and the rewards mainly benefit early game

src/main/java/com/hbm/world/feature/GlyphidHive.java
70000hp: Good idea

src/main/java/com/hbm/world/feature/GlyphidHive.java
70000hp: I didnt make LootGenerator though, and I'd rather not change it since other structures use it

src/main/java/com/hbm/world/gen/component/OfficeFeatures.java
70000hp: How did that end up there?

src/main/java/com/hbm/world/gen/component/Component.java
```diff
@@ -663,7 +673,44 @@ protected void fillWithBlocks(World world, StructureBoundingBox box, int minX, i
...
+				int posZ = getZWithOffset(x, z);
+
+				if(posX >= box.minX && posX <= box.maxX && posZ >= box.minZ && posZ <= box.maxZ) {
+
+					for(int y = minY; y <= maxY; y++) {
+						int posY = getYWithOffset(y);
+
+						if (rand.nextInt(1000) <= 8 + (pollutedLocations/10) - maxSize * 2 && hives.size() < maxSize) {
+                            if(hives.size() < 3 || hives.get(hives.size() - 1).squareDistanceTo(hives.get(hives.size() - 2)) <= 100) {
+								GlyphidHive.generateBigOrb(world, posX, posY, posZ, world.rand);
```
HbmMods, line 704: This seems like a bad idea, all these structures use the more elegant component system which does checks for bounding boxes and prevents things like cascading chunk lag, and you've bashed the old "stupid" system into it which just places blocks in an area.

src/main/java/com/hbm/world/gen/component/Component.java
```diff
@@ -663,7 +673,44 @@ protected void fillWithBlocks(World world, StructureBoundingBox box, int minX, i
...
 	}
-	
+	/** Have you ever wanted Glyphids in your building? No? Too bad! **/
+	protected void infest(World world, StructureBoundingBox box, int minX, int minY, int minZ, int maxX, int maxY, int maxZ, int maxSize, Random rand){
+		if(!enableInfestation)
+			return;
+
+		PollutionHandler.PollutionPerWorld ppw = PollutionHandler.perWorld.get(world);
+
+		int pollutedLocations = ppw == null ? 0 : ppw.pollution.size();
```
HbmMods, line 683: Treating the pollution system as a global thing doesn't really make sense since pollution doesn't even spread this far. Plus having it work like that is going to disproportionally affect servers, as there would be multiple bases with more pollution sources, even if they are really spread out.

70000hp: Hmm, fair enough, this thing alone caused a lot of headache when testing

src/main/java/com/hbm/world/gen/component/Component.java
Vaern: glyphids are a force to be reckoned with early-game, and placing them directly in buildings only discourages their exploration until later (which kinda defeats the point).
martin's also right, but a big part of the 'new' structure system was to get rid of this kind of messy, sometimes cascade-causing worldgen

70000hp: I made the base infested building chance somewhat low, I didn't want to discourage their exploration, I wanted those buildings to encourage players to make early game guns which are often overlooked and gear and raid the buildings.
I made sure the payout was fairly worthwhile for an early game player

HbmMods: > I wanted those buildings to encourage players to make early game guns

I really think the already common glyphid hives are reason enough.

70000hp: In hindsight, I did not fully account for what the old generation style could cause when smashed in with the more refined one. What I did account for was the cascading gen, via adding bounding box checks on the `infest` function. Structure hives are also much smaller given they are walless(to ensure glyphids could actually move through the building and out of it) so the chances of them going outside the alloted space for the structure is fairly slim

70000hp: I will remake the structure hive using the more refined system though, better safe than sorry

HbmMods: Yeah but that doesn't really fix the concept issue, I think I'm gonna pass on this one.

70000hp: > already

Right now glyphids are an almost irrelevant threat, with their hives being able to get blocked off easily by terrain, and their only source of expansion is a scout which lazily meanders around.
Even in early game they are very easy to avoid.

HbmMods: I don't want to hold people at gunpoint to fight off glyphids non-stop, and plenty of people seem to be bothered by them enough already.

70000hp: You can't please everyone, even if you put a giant neon sign saying evil bugs here, people would still die and complain

70000hp: There is a happy medium between decoration and unstoppable menace, and I worked for a fair while on finding it

Vaern: clueless

70000hp: I play tested these things for almost a month now and have been gathering feedback, it is possible to make glyphids interesting without making them pushovers or scape and run

70000hp: Which is why this is only part 1, in a upcoming series of PRs

PR #1185, opened 2023-09-04
Bomber Fixes

70000hp:
**Added**
New logic to the airstrike designator that fixes the long standing issue of bombers not loading properly

- It uses a fixed version of a mojank function to chunkload the chunk the bomber is going to spawn in, can also be used for similar issues

Turned the rancid if chains within designator into switches

### DIFF
--- a/src/main/java/com/hbm/items/tool/ItemBombCaller.java
+++ b/src/main/java/com/hbm/items/tool/ItemBombCaller.java
@@ -7,6 +7,7 @@ import com.hbm.lib.Library;
 
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
+import com.hbm.world.WorldUtil;
 import net.minecraft.creativetab.CreativeTabs;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.Item;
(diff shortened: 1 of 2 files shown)


src/main/java/com/hbm/items/tool/ItemBombCaller.java
```diff
@@ -16,101 +17,81 @@
...
+		switch (stack.getItemDamage()) {
+			case 0: list.add("Type: Carpet bombing"); break;
+			case 1: list.add("Type: Napalm"); break;
+			case 2: list.add("Type: Poison gas"); break;
+			case 3: list.add("Type: Agent orange"); break;
+			case 4: list.add("Type: Atomic bomb"); break;
+			case 5: list.add("Type: VT stinger rockets"); break;
+			case 6: list.add("Type: PIP OH GOD"); break;
+			case 7: list.add("Type: Cloud the cloud oh god the cloud"); break;
+			default: list.add("Type: INVALID, Report it to mod creator");
```
HbmMods, line 40: Weird default case, I don't need people to report to me when they /give themselves airstrikes with weird meta.

src/main/java/com/hbm/world/WorldUtil.java
```diff
@@ -31,6 +34,38 @@ public static void syncBiomeChange(World world, Chunk chunk) {
 		PacketDispatcher.wrapper.sendToAllAround(new BiomeSyncPacket(coord.chunkXPos, coord.chunkZPos, chunk.getBiomeArray()), new TargetPoint(world.provider.dimensionId, coord.getCenterXPos(), 128, coord.getCenterZPosition() /* who named you? */, 1024D));
 	}
 
+    /**Chunkloads the chunk the entity is going to spawn in and then spawns it
+	 * @param entity The entity to be spawned**/
+
+	/*fun fact: this is based off of joinEntityInSurroundings in World
+	  however, since mojang is staffed by field mice, that function is client side only and half-baked
+	 */
+	public static void loadAndSpawnEntityInWorld(Entity entity) {
```
HbmMods, line 43: If this code was clientside only, are you sure replicating it is going to work on the server? Did you check if any of the called functions are also clientside only or behave different on local worlds, and did you test it on a server?

70000hp: I checked all the code, there was literally no reason whatsoever for any of it to be on client-side, all it did was get a 2x2 chunk area, so it'd load, and then spawn the entity (which it didn't even do properly), I also have a server (thanks pecora's SMP) running my fork, and so far they've had no issues with the airstrikes.

src/main/java/com/hbm/items/tool/ItemBombCaller.java
70000hp: Some failsafe didn't seem like a bad idea in retrospect, incase some airstrike is added

PR #1192, opened 2023-09-08
The Goodening of Cobalt & Starmetal Gear

70000hp:
Starmetal and cobalt gear currently have recipes that would only be viable if cobalt was easy to make in early game, which it isnt, so:
**Added**
- Increased the durability and enchantability of starmetal and decorated cobalt tools 

   - The enchantability changes are good enough to convince people to actually enchant for once, which also making these tools more viable
- Decorated cobalt tools are cheaper, made by a blast furnace recipe with the cobalt tool equivalent + a meteorite ingot
- Slightly changed Starmetal tool recipes to use less cobalt and more starmetal
- Decreased the cost of cobalt armor significantly
- Increased the resistance of cobalt armor to be slightly better than AA 
- Increased the enchantability of cobalt armor

### DIFF
--- a/src/main/java/com/hbm/inventory/recipes/BlastFurnaceRecipes.java
+++ b/src/main/java/com/hbm/inventory/recipes/BlastFurnaceRecipes.java
@@ -64,12 +64,6 @@ public class BlastFurnaceRecipes extends SerializableRecipe {
 		addRecipe(ModItems.meteorite_sword_hardened,	CO,												new ItemStack(ModItems.meteorite_sword_alloyed));
 		addRecipe(ModBlocks.block_meteor,				CO,												new ItemStack(ModItems.ingot_meteorite));
 
-		addRecipe(ModItems.cobalt_sword, ModItems.ingot_meteorite, new ItemStack(ModItems.cobalt_decorated_sword));
-		addRecipe(ModItems.cobalt_pickaxe, ModItems.ingot_meteorite, new ItemStack(ModItems.cobalt_decorated_pickaxe));
-		addRecipe(ModItems.cobalt_axe, ModItems.ingot_meteorite, new ItemStack(ModItems.cobalt_decorated_axe));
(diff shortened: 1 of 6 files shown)


src/main/java/com/hbm/inventory/recipes/BlastFurnaceRecipes.java
```diff
@@ -63,7 +63,13 @@ public void registerDefaults() {
 		addRecipe(CO,									ModBlocks.block_meteor,							new ItemStack(ModItems.ingot_meteorite));
 		addRecipe(ModItems.meteorite_sword_hardened,	CO,												new ItemStack(ModItems.meteorite_sword_alloyed));
 		addRecipe(ModBlocks.block_meteor,				CO,												new ItemStack(ModItems.ingot_meteorite));
-		
+
+		addRecipe(ModItems.cobalt_sword, ModItems.ingot_meteorite, new ItemStack(ModItems.cobalt_decorated_sword));
```
HbmMods: Better idea: Make an anvil smithing recipe with heated meteorite instead of shoving tools with wood handles into a miniature crucible

HbmMods (changes requested):

src/main/java/com/hbm/inventory/recipes/BlastFurnaceRecipes.java
70000hp: Sure, forgot about the wood par

JameH2: The arirals are coming for you

70000hp: > Better idea: Make an anvil smithing recipe with heated meteorite instead of shoving tools with wood handles into a miniature crucible

image
there we go, thanks for reminding me about combustion

PR #1193, opened 2023-09-09
Inverted diode and 1.12.2's energy storages + infitity water tanks.

Toshayo:
- Port of 1.12.2 energy storage like desh, saturnite and others. I've adjusted a little capacity of dinetronium to fit new storages better. As there is no radspice in 1.7.10 I've replaced it with more verticium caus it's green :)
- Port of 1.12.2 Mk.3 and Mk.4 infinity water tanks
- Added inverted diode. The difference is, like the name implies, inputs and outputs are inverted allowing players to divide production and consumption networks without spacing too much energy storages.


LePeeperSauvage:
Please consider : never Pull Requesting anything in this repo ever again.

JameH2:
due to the existence of capacitors, i dont see how the diode makes sense

70000hp:
The former two are awful and tarnish gods creation, the latter is just pointless

JameH2:
Alright so to be fair let me explain why we believe the way we believe about this pr

1. The new battery blocks dont make sense, we have so many energy storage blocks that it really doesnt matter and all it does is serve to be bloat in the mod (and this is coming from me) 
2. The heavy infinite water tank already does the job effectively, you can make more if you need to and it really doesnt need more item variants of it but thats my two cents
3. From what im getting at, the inverted diode is just a diode flipped around? Regardless, diodes are fine as is and with the existence of capacitors. You dont really need these new diode types anyway. 

The reason for the reaction to this pr is because some of these items and stuff just serve as colorful candy bloat, that do nothing more but make people pleased because they think its something they needed when in reality the (1.7) branch can already fulfill that without having to add another existing block for the sake of looking like its more full of “diverse content” 

I hope you understand

Toshayo:
I see. Though I believe that the gap between schrabidium storage and dinetronium is rather big (in resource means). What do you thing about euphemium storage (and remove all others)?
After thinking over I agree with the fact that infinite water tanks being unnecessary if we correctly set up low-pressure steam processing.
Considering the low schrabidium capacity we need several of them to store rbmk generated power. I usually place these storages set in in/out mode separated with diodes. It makes the energy flow in one direction instead of making loops. Inverted diode would let to make this setup more compact. However, as capacitors are rather new I've not learned how do they work exactly. As far as I remember they act like a diode but with bigger storage (but much lower than energy storages) don't they?

JameH2:
capacitors are basically energy storages that act like a diode yes, by splitting i/o modes where the capacitor can only take power and when connected to the capacitor bus, the bus itself outputs power, allowing you to plug your capacitors to a generator and the output to whatever machines needed. The old method was having a battery have I/O mode but that would result in issues.

as for storage capacity, i dont see the importance of any "gap" or whatever, simply make more energy storage blocks? capacitors may have less total storage but thats made up for the resource cost and the amount you can place in the arrays that you make of them

Toshayo:
You will need 250 best capacitors to make one schrabidium storage.
Here are same capacity but one with normal diodes, second with inverted. Though using normal diodes looks maybe better and simpler to understand

JameH2:
wait are the inverted diodes the same thing as the capacitor bus?

Toshayo:
Capacitor bus outputs only on 1 side and requires a capacitor on top that has a transfer limit. Inverted diode has 1 input and 5 outputs and are connected to normal cables. Normal diode has 5 inputs and 1 output.

JameH2:
the only transfer limit of the bus is the throughout of the capacitors

Toshayo:
Commit https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/commit/37eb69e47d60ebd3d56a98a13d57c50ab3ae9b52 that adds a better capacitor makes this PR even more useless XD

PR #1195, opened 2023-09-12
Default duct sounds and versatile ModSoundType

MartinTheDragon:
- Placing default fluid ducts now plays the pipe sound as well
- Made a more versatile version of `ModSoundType`
    - Easy control over block sound subtypes
    - Allows copying from existing sounds
    - Volume and pitch envelopes
- Moved custom sound types into their own class


src/main/java/com/hbm/main/MainRegistry.java
MartinTheDragon, line 1: mfw 100 lines of import changes
if it's an issue, i can revert that
it did kinda make them nicer tho, so...

PR #1221, opened 2023-10-13
Remove NBT of probability from custom machine output items

sdddddf80:
Remove NBT of probability from custom machine output items.


HbmMods:
Simply removing the display name tag probably won't work, since it leaves behind the empty `NBTTagCompound` instance which will most likely prevent stacking, nor does it fix the underlying issue, that being the NEI handler using the original item instances from the recipes which is what causes the tags to persist to begin with. The issue has been resolved in the latest commit.

PR #1242, opened 2023-11-06
The Convenience Store

JameH2:
Gibble.

This PR adds some QOL such as being able to set fluid tanks with ID's without opening the gui, (saves as much seconds as typing "u") as well as being able to pick the same pipe in creative with no extra searching needed or whatever

enjoy!


src/main/java/com/hbm/blocks/network/FluidDuctStandard.java
JameH2, line 1: Woops i left some commented shit in, sorry

src/main/java/com/hbm/blocks/network/FluidDuctStandard.java
```diff
@@ -70,7 +75,42 @@ public void getSubBlocks(Item item, CreativeTabs tab, List list) {
 	public int damageDropped(int meta) {
 		return rectify(meta);
 	}
+	/*
```
70000hp, line 78: That is a lot of commented code, what happened here?

JameH2: The horrible attempt at the creative convenience, you can call it “mk1”


also i got rushed and i really wanted to pr this i am sorry bob

PR #1247, opened 2023-11-19
Make the research reactor compatible with IC2's water changes

Ice-Arrow:
IC2 lowers the blast resistance of water to less than 100 which breaks the research reactor's check for radiation shielding blocks making it spew radiation even when submerged in water.

With this PR the reactor will explicitly check for water blocks instead of relying on the blast resistance of water being 100 or higher.


70000hp:
The million dollar question is
Why would you even play ic2 with NTM?
Both progressions are entirely separate, and NTM already overrides practically all of ic2's material and energy requirements

Ice-Arrow:
What can I say, I just like to channel my inner Alcater and cause redundancy from time to time

In all seriousness though it's because I play multiplayer with friends who aren't so NTM inclined so the kitchen sink style pack we play includes both mods
Based on a quick search in the discord I don't think I'm the only IC2 sinner to have this problem either

HbmMods:
IC2 isn't the only mod that alters blast resistances, and plenty of people have reported issues with reactor leakage despite being underwater. Having water as an explicit radiation shielding block is a good idea.

laurynasl:
But what about river water, sea water and especially swamp water from GregTech 6 as shielding materials? ;)

HbmMods:
If the many waters of Greg have sufficient blast resistance (like vanilla water) then they will work.

PR #1265, opened 2023-12-09
The Entire Glyphid Update

70000hp:
*Added:**
Reworked considerable amounts of glyphid related content

 Enhanced pathfinding and AI:

-   Glyphids now use a task system to coordinate between various actions, those being:

1. Calling for reinforcements to a spefic location
2. Expanding the hive (used only for the scout and Big Man Johnson)

- > Scouts are also significantly better at finding a suitable hive location, and pathfind directly to it instead of meandering about

- > They spawn much less often to compensate

3. Retreating home and calling reinforcements
4. Pathfinding to a spefic location
5. Terraforming, which is when scouts direct big man johnsons to explode in certain locations, which clears off obstacles, helping with pathfinding

Reworked glyphid armor damage mechanics, making them signficantly less spongy
Nerfed fire damage against glyphids significantly, it still bypasses armor, but flamethrowers will not one-shot them
Glyphids now take 1.5x damage when afflicted by WP, to compensate for the fire nerf, and 2x damage from melee, if you've got the courage to get near them
Glyphids now flee from lanterns instead of becoming paralyzed by them

Rampant mode:

-  Changes glyphid behavior and spawning to be more aggressive, changes include:

   1. Glyphid Scouts will naturally spawn alongside normal mobs if soot levels are above a certain threshold
   2. Glyphids will always have the extended targetting enabled
   3. Glyphids can dig to waypoints, their target locations
   4. The Glyphids will expand always toward your base
   5. Scouts will spawn from the start, making glyphids start expanding off the bat
   6. Smokestacks have reduced efficiency, only reducing soot by 40%

-  All those changes are configurable

Disperser Canisters

- Act as fluid grenades, dispersing a mist of the fluid contained within them, which has unique behavior depending on the fluid (shrimply EntityMist)
- Require 2000mb of fluid to fill
- Glyphid glands, effectively the same as disperser canisters, are dropped by Glyphid Behemoths and Brendas, which hold 4000mb of sulfuric acid, and a new Pheromone fluid
>- Can be liquefied into biogas when empty
- Glyphid Scouts now have a bite dealing poison 4, to encourage the player to use ranged against them.

Pheromone:

- Gives glyphids a myriad of buffs, and a mist of it is released on death
- Entities other than glyphids get a small amount of the buffs released, giving a risky way to quickly obtain buffs to the player
- Can be refined into a concoction named Modified Pheromone, which gives the full extent of buffs to players, acting as a dispersable potion

Misc changes:
- Behemoth behavior has been reworked entirely, they now act like DRG's preatorians, with an acidic breath with a low range but high damage, and quickly goes through armor durability
>-- However, they cannot turn, or move when the attack occurs, allowing the player to dodge it
- The Spas normal fire now has the headshot bonus all other 12 gauges have (entirely unrelated, was just making me angry for a bit)
- A recipe to process glyphid meat was added, requires 1000mb of sulfuric acid and three meats, and yields 4 sulfur, 3 niter and 250mb salient green


src/main/java/com/hbm/inventory/fluid/Fluids.java
```diff
@@ -182,7 +185,8 @@ public class Fluids {
 	public static final FT_NoID NOID = new FT_NoID();
 	public static final FT_Delicious DELICIOUS = new FT_Delicious();
 	public static final FT_Leaded LEADED = new FT_Leaded();
-	
+
+	public static final FT_NoDispersable NO_DISPERSE= new FT_NoDispersable();
```
HbmMods: Reminder, that's what the VISCOUS trait is for

src/main/resources/assets/hbm/lang/en_US.lang
```diff
@@ -846,6 +847,7 @@ death.attack.mku=%1$s died from unknown causes.
 death.attack.monoxide=%1$s forgot to change the batteries in their carbon monoxide detector.
 death.attack.mudPoisoning=%1$s died in poisonous mud.
 death.attack.nuclearBlast=%1$s was blown away by a nuclear explosion.
+death.attack.nitan=%1$s was consumed by taint.
```
HbmMods: Pretty sure that's not a thing

src/main/resources/assets/hbm/lang/en_US.lang
```diff
@@ -1404,6 +1408,10 @@ hbmfluid.woodoil=Wood Oil
 hbmfluid.xenon=Xenon Gas
 hbmfluid.xpjuice=Experience Juice
 hbmfluid.xylene=BTX
+hbmfluid.caulk=Quick-Hardening Caulk
```
HbmMods: Neither is this

src/main/resources/assets/hbm/lang/en_US.lang
```diff
@@ -1580,6 +1588,9 @@ item.ammo_44_star.name=.44 Magnum Bullet (Starmetal)
 item.ammo_45.name=.45 ACP Bullet
 item.ammo_45_ap.name=.45 ACP Bullet (AP)
 item.ammo_45_du.name=.45 ACP Bullet (DU)
+item.ammo_45_drum.name=.45 ACP 100 Round Drum
```
HbmMods: Nor this

HbmMods:
Just a few small things left to do, and make sure you read the [contribution guidelines](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/master/CONTRIBUTING.md).

src/main/resources/assets/hbm/lang/en_US.lang
70000hp: Oh, oop
Those are lang entries for fork features of mine

src/main/java/com/hbm/inventory/fluid/Fluids.java
70000hp: The idea of making antimatter viscous just didn't sit right
Plus, doesn't viscous also change chemthrower behavior?

HbmMods: On main, `FT_Viscous` isn't in use anywhere as it was only meant to mark fluids that can't be turned into a mist or cloud, a feature which doesn't exist yet. As for antimatter, simply add a second exception for the `FT_Amat` trait.

70000hp:
There we go, removed the unused lang, and the No_Disperse tag

src/main/java/com/hbm/entity/mob/EntityGlyphid.java
```diff
@@ -61,102 +107,241 @@ protected void applyEntityAttributes() {
...
+							taskY = (int) taskWaypoint.posY;
+							taskZ = (int) taskWaypoint.posZ;
+
+							if (taskWaypoint.highPriority) {
+								setTarget(taskWaypoint);
+							}
+
+						}
+
+						if (taskX != none) {
```
HbmMods, line 206: wait, shouldn't this be `currentTask`?

HbmMods:
Just a couple more question before I process the merge branch further:
- Why does the `rampantTargetSetter` rely on beds? Wouldn't that mean that one could lure bugs into nowhere by sleeping once far away from base and then never again? and if nobody has slept yet, wouldn't the target be 0/0/0?
- Why do scouts get seemingly random potion buffs based on their task state?
- What made you think that inserting the new fluid definitions anywhere other than the bottom was a good idea? The disclaimer literally spans 12 lines.
- If the home of a glyphid is destroyed and the home point is mid-air, wouldn't that mean that the glyphid's pathfinding would break because of that? Is there anything that validates home points?

PR #1267, opened 2023-12-10
FEL sound

0xaa55h:
FEL should now make a sound like in 1.12 does
*is fel a hypixel skyblock reference?*


OlorinIV:
> is fel a hypixel skyblock reference?

Sadly not, FEL stands for Free Electron Laser... Just a laser-emitting device thingy.

PR #1280, opened 2023-12-19
Fix inverted door redstone functionality

Voxelstice:
a small pr. is there anything else i need to explain?

changes:
- NTM doors now have redstone functionality just like vanilla doors. (i did not add any functionality, i wrote this when i was tired)
	- why? stuff like pressure plates or motion sensors (from some mod) when activated will just close the door instead when it should do the opposite


TheVitya2127:
I liked the doors because of their simplicity. Redstone will only further force players to break them or not use them at all

JameH2:
he meant he *added* functionality, not changed it by adding it and then removed it, 

besides it doesnt take that much effort to place a button/pressure plates

PR #1288, opened 2023-12-30
More custom machine options

FOlkvangrField:
Add more custom machine options. Now custom machines can receive neutron flux and heat through the "Neutron Flux Receiver" and "Heat receiver" blocks within the structure. The recipe has also added corresponding configurable neutron flux and heat requirements, and can also be configured to generate 
or absorb pollution. (Due to the fact that hbmCustomimMachines.json under hbmConfig does not automatically refresh, please delete the old file before testing to prevent it from crashing due to reading non-existent values during loading)
image


FOlkvangrField:
@HbmMods I have noticed that this pull request has been silent for almost a month, without any comments or closure. I am not sure if this is due to any serious programming errors in this pull request, or if its changes make you feel unreasonable and speechless, or if it conflicts with your future plans for the mod, or perhaps simply because you have invested a lot of effort in developing new content and overlooked this pull request. Could you please give me a reply expressing your opinion?

HbmMods:
Sorry for the late response, I don't check the PR board all that often. The changes seem fine for the most part, but the configs absolutely have to be backwards-compatible. Old configs should never crash the game, instead the game should fill in the missing parts with default values.

FOlkvangrField:
> Sorry for the late response, I don't check the PR board all that often. The changes seem fine for the most part, but the configs absolutely have to be backwards-compatible. Old configs should never crash the game, instead the game should fill in the missing parts with default values.

Saying "crash" may be a bit exaggerated, but in fact, it is the same situation as in old configs where something was accidentally omitted or formatting errors occurred due to negligence. As a result, the custom machine will not be registered, leaving only an invalid machine control block that cannot be placed. I specifically raised this point because there may already be players using the modified config, and if they use this updated version without making any changes to the config, there may be a major chaos in their saves. 
Perhaps I can add the feature of "automatically assigning default values when reading empty values in config" to the options I added, as you said? Like this:
image

PR #1290, opened 2024-01-01
Strand Caster

70000hp:
**Added:**

1. Well, the strand caster
2. Sped up crucible pouring speed by 1.5x

**Features:**
It is a faster alternative to molds that does not have a lengthy cooling delay, instead it processes metal in batches, at the cost of requiring a full water loop, similar to the assemfac and chemfac.
It also has a GUI, allowing for easy mold swapping and internal item storage
It is obtainable in fairly early game, gated only behind welded plates, however, the large batch size makes it more fitting for later large throughput setups


src/main/java/com/hbm/blocks/machine/MachineStrandCaster.java
```diff
@@ -0,0 +1,236 @@
...
+
+        if (!player.inventory.addItemStackToInventory(cast.slots[0].copy())) {
+            EntityItem item = new EntityItem(world, x + 0.5, y + this.maxY, z + 0.5, cast.slots[0].copy());
+            world.spawnEntityInWorld(item);
+        } else {
+            player.inventoryContainer.detectAndSendChanges();
+        }
+
+        cast.markDirty();
+        world.markBlockForUpdate(x, y, z);
```
HbmMods: Considering you sync all the relevant info via tile entity anyway, you don't need this bit

src/main/java/com/hbm/blocks/machine/MachineStrandCaster.java
```diff
@@ -0,0 +1,236 @@
...
+
+        cast.slots[0] = null;
+        cast.markDirty();
```
HbmMods: And you don't need this a second time

src/main/java/com/hbm/inventory/container/ContainerMachineStrandCaster.java
```diff
@@ -0,0 +1,75 @@
...
+            stack = originalStack.copy();
+
+            if (index <= 6) {
+                if (!InventoryUtil.mergeItemStack(this.inventorySlots, originalStack, 7, this.inventorySlots.size(), true)) {
+                    return null;
+                }
+
+                slot.onSlotChange(originalStack, stack);
+
+            } else if (!InventoryUtil.mergeItemStack(this.inventorySlots, originalStack, 0, 7, false)) {
```
HbmMods: Why should the player be able to shiftclick things back into the output?

src/main/java/com/hbm/inventory/container/ContainerMachineStrandCaster.java
```diff
@@ -0,0 +1,75 @@
...
+    public ContainerMachineStrandCaster(InventoryPlayer invPlayer, TileEntityMachineStrandCaster caster) {
+        this.caster = caster;
+
+        //the wretched mold
+        this.addSlotToContainer(new SlotNonRetarded(this.caster, 0, 57, 62));
+
+        //output
+        for (int i = 0; i < 3; i++) {
+            for (int j = 0; j < 2; j++) {
+                this.addSlotToContainer(new SlotNonRetarded(this.caster, j + i * 2 + 1, 125 + j * 18, 26 + i * 18));
```
HbmMods: You probably want to use `SlotMachineOutput` for this one

src/main/java/com/hbm/main/ResourceManager.java
```diff
@@ -50,7 +50,10 @@ public class ResourceManager {
 	public static final IModelCustom boiler_burst = new HFRWavefrontObject(new ResourceLocation(RefStrings.MODID, "models/machines/boiler_burst.obj"));
 	public static final IModelCustom boiler_industrial = new HFRWavefrontObject(new ResourceLocation(RefStrings.MODID, "models/machines/industrial_boiler.obj"));
 	public static final IModelCustom hephaestus = new HFRWavefrontObject(new ResourceLocation(RefStrings.MODID, "models/machines/hephaestus.obj"));
-	
+
+	//Caster o' Strands
+	public static final IModelCustom strand_caster = AdvancedModelLoader.loadModel(new ResourceLocation(RefStrings.MODID, "models/machines/strand_caster.obj"));
```
HbmMods: Anything with smooth shading needs to be loaded as `HFRWavefrontObject`

src/main/java/com/hbm/tileentity/machine/TileEntityFoundryBasin.java
```diff
@@ -7,7 +7,16 @@
...
 
 public class TileEntityFoundryBasin extends TileEntityFoundryCastingBase implements IRenderFoundry {
-	
+
+	public TileEntityFoundryBasin() {
+		super(2);
+	}
+
+	@Override
+	public String getName() {
```
HbmMods: Why would the foundry basin base class need that method, even though all standard basins don't even have a GUI?

HbmMods (changes requested):

src/main/java/com/hbm/blocks/machine/MachineStrandCaster.java
70000hp: Yeah I was really confused as to how the syncing works

src/main/java/com/hbm/tileentity/machine/TileEntityMachineStrandCaster.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    @Override
+    public int getCapacity() {
+        ItemMold.Mold mold = this.getInstalledMold();
+        return mold == null ? 50000 : mold.getCost() * 10;
+    }
+
+    private int getWaterRequired() {
+        return getInstalledMold() != null ? 5 * getInstalledMold().getCost() : 50;
```
HbmMods, line 205: The second half of the ternary assumes `getInstalledMold()` to be null, then calls a method anyway

HbmMods (changes requested):

src/main/java/com/hbm/tileentity/machine/TileEntityMachineStrandCaster.java
70000hp: Given the water required only gets deducted if the thing can process, which in the case, it couldn't, this wouldn't actually affect anything
This is more of a failsafe

HbmMods: I'd rather if your failsafe didn't include a built-in crash

PR #1293, opened 2024-01-05
Pipettes!

BallOfEnergy1:
Added:
Pipettes

These are able to grab millibucket amounts out of machines, preventing the wasting of fluids and allowing for precision fluid crafting.

(still WIP, will confirm when ready)


src/main/java/com/hbm/items/tool/ItemPipette.java
```diff
@@ -0,0 +1,175 @@
...
+
+public class ItemPipette extends Item implements IFillableItem {
+
+    public ItemPipette() {
+        this.setMaxDamage(0);
+    }
+
+    @SideOnly(Side.CLIENT) protected IIcon overlayIcon;
+
+    public int amount = this.getMaxFill();
```
HbmMods: there's only one item instance and therefore that value is shared for each pipette, you have to use NBT for that

src/main/java/com/hbm/items/tool/ItemPipette.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public FluidType type = Fluids.NONE;
```
HbmMods: same here

src/main/java/com/hbm/items/tool/ItemPipette.java
```diff
@@ -0,0 +1,175 @@
...
+    public int getFill(ItemStack stack, FluidType type) {
+        if(!stack.hasTagCompound()) {
+            initNBT(stack);
+        }
+
+        return stack.stackTagCompound.getInteger(type.getName());
+    }
+
+    @Override
+    @SideOnly(Side.CLIENT)
```
HbmMods: pretty sure that annotation will at best not let this work on servers and at worst crash the game

src/main/java/com/hbm/items/tool/ItemPipette.java
```diff
@@ -0,0 +1,175 @@
...
+
+        if(!stack.hasTagCompound()) {
+            initNBT(stack);
+        }
+
+        if(!world.isRemote) {
+            if (this.getFill(stack, type) == 0) {
+
+                if(this != ModItems.pipette_laboratory)
+                    this.amount = player.isSneaking() ? Math.min(this.amount + 50, 1_000) : Math.max(this.amount - 50, 50);
```
HbmMods: i genuinely can't even tell what this code is supposed to do, or why you're handling the amount at all when you just checked the amount to be 0

HbmMods (changes requested):

src/main/java/com/hbm/items/tool/ItemPipette.java
BallOfEnergy1: It's meant to allow the player to change the amount of max fluid in the pipette, probably gonna rename the variable from `amount` to `maxAmount` or something of the like. It checks if the fill is 0, and if so allows the changing of the max capacity in units of either 1mB or 50mB, depending on if it's a laboratory pipette or a normal one.

BallOfEnergy1:
NBT fixed,
textures done,
tested on servers,
please review in case I missed anything, its 2 am and I've been working on this for 7 hours.

src/main/java/com/hbm/crafting/ToolRecipes.java
```diff
@@ -128,7 +128,10 @@ public static void register() {
 		CraftingManager.addRecipeAuto(new ItemStack(ModItems.bismuth_tool, 1), new Object[] { "TBT", "SRS", "SCS", 'T', TA.nugget(), 'B', ModItems.nugget_bismuth, 'S', ANY_RESISTANTALLOY.ingot(), 'R', ModItems.reacher, 'C', ModItems.circuit_aluminium });
 		CraftingManager.addRecipeAuto(new ItemStack(ModItems.sat_designator, 1), new Object[] { "RRD", "PIC", "  P", 'P', GOLD.plate(), 'R', Items.redstone, 'C', ModItems.circuit_gold, 'D', ModItems.sat_chip, 'I', GOLD.ingot() });
 		CraftingManager.addShapelessAuto(new ItemStack(ModItems.sat_relay), new Object[] { ModItems.sat_chip, ModItems.ducttape, ModItems.radar_linker });
-		
+
+		CraftingManager.addRecipeAuto(new ItemStack(ModItems.pipette, 1), new Object[] { "  L", " G ", "G  ", 'L', ModItems.ingot_biorubber, 'G', KEY_CLEARGLASS});
```
HbmMods: `ModItems.ingot_biorubber` should never be used as a recipe ingredient, it always has to be interchangeable with rubber, so use `ANY_RUBBER.ingot()` instead

src/main/java/com/hbm/inventory/recipes/AssemblerRecipes.java
```diff
@@ -696,6 +696,12 @@ private static void registerDefaults() {
 				new ComparableStack(ModItems.motor, 2),
 			}, 200);
 
+		makeRecipe(new ComparableStack(ModItems.pipette_laboratory, 1),new AStack[] {
```
HbmMods: Feels a bit tedious to make in an assembler, especially for what is essentially just a worse pipette. Maybe use the anvil instead? Also, what are the circuits for?

HbmMods (changes requested):

src/main/java/com/hbm/inventory/recipes/AssemblerRecipes.java
BallOfEnergy1: The circuits are there because the laboratory pipette is considered a form of "micropipette", which have systems/mechanisms much more complex than normal pipettes.

BallOfEnergy1:
woo!

PR #1295, opened 2024-01-09
Updated ru_RU.lang

Doctor17-git:
updated russian localization


username228a91:
crafting table in russian version of game is named верстак so i think it would be автоматический верстак instead of стол автокрафта

PR #1306, opened 2024-01-21
Thingamabob

245tt:
### Quality of life stuff:
- template folder search box focuses when opening GUI
- anvil now shows what items you don't have

### DIFF
--- a/src/main/java/com/hbm/inventory/gui/GUIAnvil.java
+++ b/src/main/java/com/hbm/inventory/gui/GUIAnvil.java
@@ -286,17 +286,19 @@ public class GUIAnvil extends GuiContainer {
 			if(stack instanceof ComparableStack)  {
 				ItemStack input = ((ComparableStack) stack).toStack();
 				boolean hasItem = false;
+				int amount =0;
 				for (int i = 0; i < inventory.mainInventory.length; i++) {
 					ItemStack stackItem = inventory.mainInventory[i];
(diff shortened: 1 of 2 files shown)


src/main/java/com/hbm/inventory/gui/GUIAnvil.java
```diff
@@ -283,16 +285,44 @@ public List<String> recipeToList(AnvilConstructionRecipe recipe) {
...
-				list.add(">" + input.stackSize + "x " + input.getDisplayName());
-				
+				boolean hasItem = false;
+				for (int i = 0; i < inventory.mainInventory.length; i++) {
+					ItemStack stackItem = inventory.mainInventory[i];
+					if(stackItem == null)
+					{
+						continue;
+					}
+					if (stackItem.getItem() == input.getItem() && stackItem.stackSize >= stack.stacksize) {
```
HbmMods: This won't work for a lot of cases. For one, it doesn't check for item metadata so things like subitems (like different cast plate materials) would all trigger the check even though they aren't accepted by the actual recipe. In addition, it won't ever trigger for split stacks, two stacks of 32 would not satisfy the requirement for 64 items, even though it would work in the actual recipe. Consequently, a requirement of anything higher than 64 would never work because the player can't carry stacks that big

245tt:
this should work now

PR #1314, opened 2024-01-25
Fuel upgrade

Technomirfuture132:
-----Fuel upgrade-----
Have been added:
New types of fuels are bio-oil, synthetic varieties of hydrocarbon fuels.Bio-oil is a complex of oxygen-containing compounds and valuable chemical raw materials containing hydrocarbons.When hydrotreating bio-oil, its combustible properties are improved and its corrosion is neutralized.

Brine and hot brine - can be used as a liquid for a heat exchanger and a PWR moderator.

The Russian localization has been updated, but I'm still in the process, and I'm following the updates.
Several new recipes for the chemical plant have been added.

A new recipe for the production of chlorine and sodium by electrolysis of brine has been added.
Solidification recipes have been added for new fuels, taking into account their caloric content.

 The new items are salt, mineral dust, and their miniature versions, as well as an iron-cobalt catalyst.Mineral dust is useless for now, but in the future I will add new recipes for its centrifugation.

A new source of geothermal energy has been added for the geothermal heat exchanger - radioactive lava.It is 100 thermal units hotter than ordinary volcanic lava.
A new method of producing petroleum coke from bio-oil.


Ashen-sysadmin:
Is the radioactive lava the new one that HBM added recently?

Technomirfuture132:
> Is the radioactive lava the new one that HBM added recently?

Yes, this is a new type of lava from the new hbm update. It is assumed that the lava will be hotter and can serve as a good source of energy if you have a geothermal heat exchanger. I don't know why it wasn't added to the update right away.

Ashen-sysadmin:
Probably because it solidifies too quickly. You don't have flowing rad lava for long.

Technomirfuture132:
> Probably because it solidifies too quickly. You don't have flowing rad lava for long.

I do not know, I have tested this method of energy production in my world, if a geothermal heat exchanger is located near the mouth of a volcano, it copes with its task perfectly.If it is not a shield volcano, then the lava should form a cone-shaped mountain of blocks of slaked sellafield, and the heat exchanger works stably closer to the vent.

HbmMods:
* Salt from brine is something I've always consciously avoided - pretty much every single tech mod that has chlorine uses this chain, and it always goes water (infinite) > slightly saltier water > chlorine. I' e always found that approach boring, which is why we have like 4 other ways of making chlorine already.
* I'm not sure why brine would be a PWR coolant since it's capacity is identical to regular coolant, but it's cooling speed is 50% higher despite being a lot cheaper than coolant.
* I don't see the point in adding mineral dust when it doesn't even do anything, nor is resource extraction from water all that exciting to begin with.
* Bio oil seems in terms of acquisition pretty similar to biofuel, which makes it redundant as a fuel.
* Fractioning hot bio oil into light oil using only a fractioning tower breaks progression as it bypasses the need for a refinery for making desh.
* Having all these syntheic variants of already existing oils doesn't seem to add anything other than "having more stuff to burn", and we aren't exactly short on that.
* I'm not sure why there even has to be synthetic variants of oil made from liquefying syngas, considering crude oil can already be extracted in low amounts from coal, and bedrock oil is infinite anyway.
* Being able to crack synkerosene into reformate gas makes reformate gas available at cracked oil level, effectively bypassing the RBMK, as bismuth can be extracted from certain bedrock ores with high-performance solvent made from refgas and chlorine.

PR #1318, opened 2024-01-31
Weapon animation rework

MellowArpeggiation:
Hoo boy, what do we have here? A rework of the animation system that both brings it up to modern standards while maintaining full compatibility with animations that have already been authored? That also maintains the ability to manually add keyframes for those that don't like using animation software?

The rework is simple, instead of a bus of keyframes that takes 3 values and assigns a duration to the triplet, instead provides a 9 value bus (transform, rotation, scale) that allows for each channel to have its own frame timing. This aligns much more closely to the F-Curves system found in Blender and many other 3D animation software, without removing any functionality or capabilities. This also removes the need for separate buses for TRANSLATION and ROTATION.

See the SPAS-12 for an example that uses the new `addKeyframePosition` and `addKeyframeRotation` manual keyframe entry methods (old style), and see `ntm-animator.blend` for the process to build your own animations within Blender (complete with documentation and python scripts for JSON import/export).

On top of the rework, support for shotgun reload animations has been added, by using separate animation segments that run sequentially, see the Samuel (ks23) shotgun for a fully animated example (If nothing else, check out this weapon, it feels fantastic to use now).

This is all very complicated so I'm happy for a chat to discuss the changes or run you through any of the more complex parts!

Top level changelog:

- New animation loader, uses JSON to store keyframes and offsets
- New .blend file (Blender 4.0.2) with python scripts that implements the animation loader format
- Add animations to a bunch of guns that are probs gonna be cut (oh well, animation is _much quicker_ and easier now)
    - The .357 revolvers
    - The .44 revolvers (didn't do the reload animation, after I found out most guns are to be cut)
    - The Cursed revolver(? it's like, a pistol tho)
    - The Samuel shotgun (ks23)


HbmMods:
LIGHTWEIGHT ANIMATION LOADER?

BACKWARDS COMPATIBILITY?

SEQUENTIAL ANIMATIONS?



grafik

PR #1320, opened 2024-01-31
Dispersing artillery shells

9Valjew:
Allows crafting of dispersible canisters with express shells
Does not return the empty canister
Does not spawn lander shell if not needed
Copies the chlorine mist spawning because I cba writing that part out


src/main/java/com/hbm/crafting/handlers/CargoShellCraftingHandler.java
```diff
@@ -56,6 +56,9 @@ public ItemStack getCraftingResult(InventoryCrafting inventory) {
 				copy.stackSize = 1;
 				cargo = copy;
 			}
+			if(stack.getItem() == ModItems.disperser_canister) {
+				stack.getItem().setContainerItem(null); //removes the empty canister from the output
```
HbmMods: doing it like this will make it so that the disperser canister won't have a container item at all anymore after this recipe is used once

9Valjew:
Thats whar I thought but apprently it still keeps it (somehow)
You are free to check by all means that was implemented just as a check and
ended up working

9Valjew:
unless you mean container item as in in future crafts which makes sense
for now, there are no recipes that need a disperser canister, and i cant think of a recipe which would need one
it still works as a fluid container, but idk, its your choice

9Valjew:
(Also I have to say that in my opinion it should not have a container item by the literal dispersing nature, if I use it, for example, in a chemical warhead or whatnot, I seek the disperser quality, otherwise I could just have used 2 universal fluid tanks if I wanted 2000mb)

9Valjew:
aoogh I will try to find a workaround

9Valjew:
Ok now I:
Changed the disperser canister to have no container item because fuck it?
cleaned up the code handling the disperser canister being a container item in the cargo handler
Bobbert I am begging you I am on my knees

Vaern:
image

HbmMods:
image

the true question is why we even need an artillery shell like that at all, and why it has to be dispersers inside cargo shells, which isn't even documented, instead of having disperser shells

PR #1323, opened 2024-02-01
Custom machine name multi language support

FOlkvangrField:
Add custom machine name multi language support. By adding the field "tile. cm_‘unlocalizedName’. name=xxx" to the lang file, custom machines can be easily given multilingual names. Additionally, by adding the field "tile. cm_‘unlocalizedName’. desc=xxx" to the lang file, custom machines can be easily given multilingual, line breaking explanatory text.


HbmMods:
I'm not sure if that's a good idea since that involves tampering with the JAR file itself, and I'd rather the configuration were done in a centralized place. Ideally, the custom machine config would have an optional array with key value pairs that assign a localized name to each lang file name.

FOlkvangrField:
> I'm not sure if that's a good idea since that involves tampering with the JAR file itself, and I'd rather the configuration were done in a centralized place. Ideally, the custom machine config would have an optional array with key value pairs that assign a localized name to each lang file name.

Does "Tampering with the JAR file itself" mean that the mod packs author needs to edit the lang file within the JAR file to add or modify these texts? If so, then the mod packs author or mod packs translator can easily add and modify these texts through the resource pack without directly editing the jar file itself, just like this:
[HBMCustomMachineText.zip](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/files/14381336/HBMCustomMachineText.zip)

HbmMods:
Fair point, but I still feel like having to make a resource pack just for localization is a bit cumbersome, it would be much better if all settings related to custom machines were consolidated in the config.

HbmMods:
Localization has been added in the most recent commit, names can be added in the config file directly like this:

grafik

The locale names use the same format as the .lang files.

FOlkvangrField:
> Localization has been added in the most recent commit, names can be added in the config file directly like this:
> 
> grafik
> 
> The locale names use the same format as the .lang files.

Ok

PR #1326, opened 2024-02-06
Add fluid siphon

MellowArpeggiation:
Ever accidentally overfilled the arc welder and wanted to retrieve the erroneous fluid? Introducing the **Siphon**:

The siphon can be constructed with 3 glass and 2 rubber. If used on any machine with fluid _inputs_, it will extract the fluid into any valid containers in the players inventory (including a pipette if the remainder in the tank is less than 1 bucket).

The siphon will not extract anything into any invalid containers, so no, you can't extract absurdly hot plasma from your reactor, and no HP solvent in regular containers.

image


Ashen-sysadmin:
This is beautiful.

PR #1340, opened 2024-02-18
Old-school Super Shotgun animations

MellowArpeggiation:
Rip_and_tear-814693726

Also includes a slight change to the behaviour of worldgen mines, which prevents them from blowing up from mob spawns well before they actually pose a threat to the player (with the intent that these will instead be placed in buildings rather than everywhere, eventually). Player placed mines are unchanged.


HbmMods:
image

PR #1357, opened 2024-02-23
Fixed gamebreaking bug with the leviathan turbine

9Valjew:
I should get a bobble for this fr


JameH2:
The hell did you change

9Valjew:
> The hell did you change

tumbine

gmhs022:
Oh my god I see it, there was a 1-pixel error on the upside-down warning sign on the texture

HbmMods:
regrettably i have to inform you that we can't fix this bug because it would cause severe compatibility issues with older worlds

9Valjew:
> regrettably i have to inform you that we can't fix this bug because it would cause severe compatibility issues with older worlds

Such a shame, I guess the mod is now in shambles

PR #1364, opened 2024-02-27
Change some assembly recipies to make use of hydraulic pistons.

Ashen-sysadmin:
I've changed some key post-oil recipes to make use of the hydraulic pistons that are currently only used by conveyor upgrades and electric presses.

Namely, the compressor, assembly factory, and fracking tower all require a few hydraulic pistons, instead of or in addition to their motor costs.


HbmMods:
The other changes seem fair, if inconsequential, but why does the hydraulic fracking tower need hydraulic pistons? The addition only makes the recipe more complicated and doesn't seem to serve any gating/cost/complexity reduction purpose.

Ashen-sysadmin:
Primarily because there aren't many machines that would make sense using pistons, and I would like to increase the demand for lubricant.

I'm not committed to the particulars, and in retrospect, the assembly factory should use either all motors or all pistons, not both.

PR #1373, opened 2024-03-05
extremely minor silo change and minor fix

Vaern:
Added a lever for the hatch (as the missile would visually clip with the still-opening hatch on launch), as well as fixing a minor oversight in calculating the average height.


HbmMods:
grafik

Vaern:
artworks-uWIKBee2f39lm2CL-5YOQHA-t500x500

PR #1385, opened 2024-03-14
It was an Axe-dent I swear

burningwater202:
-Adds Bismuth Axe
-Adds Molten Axe
-Adds Chlorophyte Axe
-Adds Mese Axe

Basically the axe equivalent to the super pickaxes. 
Instead of pickaxe/shovel combo they're axe/sword combo, with the damage roughly equating to the same amount as the meteorite axe that would be available at the respective progression level

Me and Mia got tired of not having an infinite dura axe besides Desh, can you blame us?


70000hp:
jesus h christ how many trees does a mf need to cut?

burningwater202:
6 1/2, give or take

After that you're on your own

PR #1387, opened 2024-03-17
Quantum virtual particles

MellowArpeggiation:
> _I never thought I'd **see** a Resonance Cascade, let alone create one..._

This PR changes the Schottky Particle Accelerator Diode behaviour to introduce the concept of _virtual particles_. Each valid exit of a Schottky diode will now create a parallel universe virtual particle, such that the accelerator **simultaneously uses all valid path**s. Assuming no building errors are encountered on any path, it will use the lowest momentum particle that is valid for the current recipe and collapse it to become a _real particle_, consuming only the power from plugs that the real particle encountered.

If a building error is encountered, the erroring particle is made real, consuming only the power from the broken segment of the accelerator, and cancelling any successful particles. This makes it much easier to debug any building errors, since the accelerator won't ever exist in the superposition of simultaneously failing and succeeding.

The reason for this change is simple, to remove the need to dismantle your particle accelerator just to screwdriver your Schottkys every time you wish to change the accelerator power for different recipes. This has value in automating the accelerator as well!

Also included:
* Fog doesn't cache unnecessarily, improving responsiveness outside of nuclear biomes
* Fog blending should work with render distances greater than 16 now

### DIFF
--- a/src/main/java/com/hbm/main/ModEventHandlerRenderer.java
+++ b/src/main/java/com/hbm/main/ModEventHandlerRenderer.java
@@ -428,8 +428,8 @@ public class ModEventHandlerRenderer {
 		int[] ranges = ForgeModContainer.blendRanges;
 		int distance = 0;
 		
-		if(settings.fancyGraphics && settings.renderDistanceChunks >= 0 && settings.renderDistanceChunks < ranges.length) {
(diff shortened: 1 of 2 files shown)


HbmMods:
grafik

PR #1399, opened 2024-03-26
Gerald alt recipe as config option

Ashen-sysadmin:
Adds a config option to replace the current Gerald recipe with a recipe that does not need pink logs or burnt bark.

The alt recipe keeps BF warhead, CMB scrap, and the boss medals, but replaces the remaining ingredients with a mix of industrial items chosen to both be expensive and suitable for a self replicating lunar robotics platform.

### DIFF
--- a/src/main/java/com/hbm/config/GeneralConfig.java
+++ b/src/main/java/com/hbm/config/GeneralConfig.java
@@ -35,7 +35,8 @@ public class GeneralConfig {
 	public static boolean enableMOTD = true;
 	public static boolean enableGuideBook = true;
 	public static int hintPos = 0;
-
+	
+	public static boolean alternateGeraldRecipe = false;
 	public static boolean enableExpensiveMode = false;
 	
 	public static boolean enable528 = false;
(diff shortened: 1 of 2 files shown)


Ashen-sysadmin:
Goddammit, eclipse screwed up text in the config, I'm going to have to go back and fix it.

gmhs022:
Couldn't you do this with the recipe config system already in the mod?

Ashen-sysadmin:
> Couldn't you do this with the recipe config system already in the mod?

Yes, but the same could be said about any of the other config based recipe changes. This gives players handholding if they want to change the recipe, while still keeping it expensive enough to feel fair.

gmhs022:
Also this feels too cheap in areas, only overclocked circuits and with no nuclear charges or similar, not even starmetal. Crystalized Fullerite is nice though, perhaps the nuclear charges could be re-added and the overclocked circuits replaced with high-perf circuits.

Ashen-sysadmin:
Starmetal sucks to produce, and when I floated the idea of high-perf in discord, people considered it too expensive.

Vaern:
that's stupid, high perf is just a desh ingot and a schrab ingot. even the latter is easy once you get a hadron

Ashen-sysadmin:
Ask 7 about the circuits, -same with keeping the medals.

Vaern:
seven is not exactly a reliable source of information

Ashen-sysadmin:
How about this, I revert nuclear charges and starmetal, replacing the drones and assemfacs. That way, the only changes are the 2 horns, and the 2 RNG based log items.

Vaern:
👍

HbmMods:
Not sure if that's even necessary, I'm working on unfucking lategame anyway so in summer at the latest the original recipe will cease to exist anyway. Plus, assembler configs already exist, the other recipe toggles usually replace a large number of recipes at once. And finally I'm not sure if the ingredients are fitting for Gerald, the final larger thing you can even make, save for all the DFC stuff that doesn't really have a whole lot of practical use anyway. I had some ideas regarding what materials it should use, and one of them involved an expensive part you'd have to automate at a large scale, combining products of all previous automation sectors (various crucible metals, vacuum oils, syngas from coal liquefaction, maybe some general bedrock ore metals, etc). If I were to change the recipe *right now* I wouldn't even bother with a config since the current recipe is a really shitty placeholder anyway.

Ashen-sysadmin:
>Not sure if that's even necessary, I'm working on unfucking lategame anyway so in summer at the latest the original recipe will cease to exist anyway.

Good. I was just sick of hearing the complaints about pink log and bark, and wanted to spare the people who are playing right now. If I was mistaken about the amount of demand, then I'd happily close it.

> combining products of all previous automation sectors (various crucible metals, vacuum oils, syngas from coal liquefaction, maybe some general bedrock ore metals, etc). 

That was the initial goal, but everyone else talked me out of it. The lenses are the main thing that I think should really be included, because they force heavy industry for TATB automation.

PR #1407, opened 2024-03-29
Config option to disable cooling tower particle emissions

CM436:
All this does is add a toggle in the config to disable the steam particles from cooling towers even on "All" and "Decreased" particle settings. I only thought of this because I have like 10 cooling towers around my base and it's hard to see anything without changing the settings to minimal but I'd like to keep the rest of the particle effects on.

There's also 6 commits because I made a couple mistakes and each time I noticed one I'd just make a new commit cause I assumed that it would be the last change (I'm still new to this stuff).

### DIFF
--- a/src/main/java/com/hbm/config/GeneralConfig.java
+++ b/src/main/java/com/hbm/config/GeneralConfig.java
@@ -34,6 +34,7 @@ public class GeneralConfig {
 	public static boolean enableFluidContainerCompat = true;
 	public static boolean enableMOTD = true;
 	public static boolean enableGuideBook = true;
+	public static boolean enableSteamParticles = true;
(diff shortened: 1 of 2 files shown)


gmhs022:
You should probably try using a high power condenser, which is equivalent to 10 cooling towers, not against the config option tho, could be useful.

CM436:
that is a good idea, although the cooling towers are across multiple reactors each only needing 1 or 2 so it would use a lot of power to replace them all with powered condensers

gmhs022:
The power usage is based on how much LPS is condensed, so it'd take the same amount no matter how split up it is, the total energy cost isn't that bad either, it's a fraction of the power generated from any setup requiring one.

PR #1415, opened 2024-04-01
bob i swear to god

70000hp:
natural scout spawning has been broken ever since you added the light check.
fixed now


Vaern:
aired

PR #1426, opened 2024-04-09
QoL changes

MellowArpeggiation:
qol

Includes a few QoL changes, I have more I want to do with the Geiger counters, but I don't currently have the time.

Changes included are as follows:
- Multi Fluid Identifier search box is automatically focused when opened
- Structure/Dungeon config option now has an option to respect world `Generate Structures` flag, either can be set to `true`, `false`, or `flag`
- Geiger counters integrated into armor now measure audible counts _after_ protection is calculated
   - Regular counters are unchanged, since they aren't under the armor and would count as normal

Fixes:
- Fluid compressing could be skipped by putting fluid containers into the slots
   - For the chemplant, it'll just ignore the containers now
   - For the hydrotreater and vacuum refinery, the GUI has been modified to show the pressurization requirement, and the slots are deprecated (you can retrieve items from them, but they otherwise can't be interacted with)


Vaern:
luv that first change

PR #1445, opened 2024-04-18
Replace DisplayLists with VBOs (Improved Angelica support)

MellowArpeggiation:
This PR not only replaces the DisplayList implementation with more modern and fast VBOs, it also fixes up most of the extant compatibility issues with Angelica, including:
* Connected texture blocks now render correctly
* Machines with 3D models now render correctly

Known Issues (with Angelica):
* Reeds render incorrectly in deep water, turn off _Video Settings_ > _Advanced_ > _Use Compact Vertex Format_ to fix.

### DIFF
--- a/src/main/java/com/hbm/render/block/ct/RenderBlocksCT.java
+++ b/src/main/java/com/hbm/render/block/ct/RenderBlocksCT.java
@@ -26,11 +26,8 @@ public class RenderBlocksCT extends RenderBlocks {
 	VertInfo bc;
 	VertInfo br;
 	
-	Tessellator tess;
-	
 	public RenderBlocksCT() {
 		super();
-		this.tess = Tessellator.instance;
 	}
 	
 	public void prepWorld(IBlockAccess acc) {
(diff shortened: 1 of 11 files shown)


src/main/java/com/hbm/render/block/ct/RenderBlocksCT.java
```diff
@@ -315,32 +312,32 @@ public void renderBlockAsItem(Block block, int meta, float mult) {
 		GL11.glRotatef(90.0F, 0.0F, 1.0F, 0.0F);
 		GL11.glTranslatef(-0.5F, -0.5F, -0.5F);
 		
-		tess.startDrawingQuads();
-		tess.setNormal(0.0F, -1.0F, 0.0F);
+		Tessellator.instance.startDrawingQuads();
```
mitchej123, line 315: Small nit - it would be slightly faster to store the tessellator in a local var if accessing Tesselator.instance multile times; otherwise angelica is running through and replacing each of these with a .get() which might include a ThreadLocal lookup.

PR #1455, opened 2024-04-23
i had to switch gears on em

Vaern:
y'know, swerve, left-turn, steering wheel on em
 - Added Redstone-Over-Radio Logic Receiver
   - Maps input signal to a specific redstone signal by evaluating conditions in a set order, with the first applicable condition determining the strength of signal
     - Compares strings and numbers to constants
     - 10 operators: <, <=, >=, >, =, !=, matches, does not match, contains, does not contain
       - ui doesn't suck ass because you can scroll instead of clicking nine times
     - Can switch between ascending and descending order
  - Allows the parsing of signals from item counters and more complex parsing in general (e.g. strings from transmitters)
  - blinks


HbmMods:
grafik

PR #1459, opened 2024-04-24
Switch plutonium for neptunium from schraranium processing

Vaern:
CoC
 - Replaced the generic plutonium output from schraranium processing recipes with neptunium
   - should allow much easier accumulation of neptunium for normally annoying recipes, such as schrabidium batteries
   - more importantly, gives a reason to make the lower tier of schrab fuels


70000hp:
hough.

Ashen-sysadmin:
One issue, is that this makes certain plutonium isotopes a lot harder to get.
-Balancewise, it means we should make the breeder reactor more usable.

Vaern:
Not really, the cyclotron can produce generic plutonium from neptunium.

gmhs022:
Based, and I never really found generic plutonium all that useful anyways.

PR #1461, opened 2024-04-26
Add custom recipe system for gas centrifuge and SILEX

FOlkvangrField:
Add custom recipe system for gas centrifuge and SILEX , and made some changes to the recipe processing logic and nei support for gas centrifuges. Now we can create a new gas centrifuge formula process, configure whether a certain level of PseudoFluid needs overclocking upgrade, whether additional items will be produced if there is no overclocking upgrade, whether it can idle, and whether it can idle to consume three times the amount of the next level of PseudoFluid produced by one centrifugation of PseudoFluid to output additional items. I have mostly written the specific details in the comment at the top of the JSON file. Regarding Nei support, I have written a lot of shits to ensure readability as much as possible. Simply put, it will generate a recipe for Nei to read for a certain level of PseudoFluid in a processing flow, including whether to add overclocking or idle. However, it still looks a bit rough, and I haven't found a better implementation method yet.
image

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,7 +11,7 @@ credits=HbMinecraft,\
   \ Alcater (GUI textures, porting),\
   \ MellowArpeggiation (new animation system, turbine sounds, sound fixes, industrial lights, better particle diodes),\
   \ Pheo (textures, various machines, models, weapons),\
-  \ Vær (gas centrifuges, better worldgen, ZIRNOX, CP-1 parts, starter guide),\
+  \ Vï¿½r (gas centrifuges, better worldgen, ZIRNOX, CP-1 parts, starter guide),\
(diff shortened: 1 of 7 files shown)


FOlkvangrField:
So @HbmMods , how do you think about it? I've thought about displaying the entire process of a fluid directly on one of the NEI interfaces, including branching processes with or without overclocking and idling, but then I would have to limit the number of steps in a set of processes to avoid going out of the display range

HbmMods:
I haven't had the time to review the PR yet, the gas centrifuge code is quite delicate and config PRs have a history of being broken in some way so I'd rather take my time for this one.

FOlkvangrField:
OK, looking forward to your achievement.

MurdoqZodoq (approved):

PR #1463, opened 2024-04-28
Small tweaks

70000hp:
**Added**

-  Small set of anvil scrapping recipes, mainly for antenna parts, but also the stirlings, due to how quickly they become obsolete
- Pickaxes and Terra Drills now can mine glass, glowstone, and ice, and use veinminer on them


JengaG1:
uhh based pull request

gmhs022:
Thank god for the glowstone change that was always annoying as hell to gather.

PR #1465, opened 2024-04-30
Fix quotes in resource file

Majora320:


70000hp:
What
Why?

9Valjew:
huh?

JameH2:
Wdym fixing? Those slashes are escape sequences

Majora320:
Open Minecraft and look at the descriptions of the items I changed, the backslashes get rendered verbatim.

MellowArpeggiation:
He's right, the backslashes aren't required in `.lang` files, and are rendered as regular text. It's likely that the text was copied from code (where the escape sequence _is_ required) when migrated to the localized files with the backslashes still intact.

Vaern:
this is why you localize from the start

PR #1478, opened 2024-05-07
Improve RBMK render performance (and explosion rework whoops)

MellowArpeggiation:
Switch TESR RBMK rendering to VBOs, leaving the regular ISBRHs as is.
Also includes a slightly improved wiki renderer and fixed shallow foundry basin rendering.

Also includes a minor explosive block rework because I accidentally didn't branch out before committing:
* All blocks now spawn entities when destroyed by an explosion, rather than instantly exploding
* Explosive/Waste barrels now fly off as an entity, exploding when contacting a surface
* All other explosives instead spawn a zero tick entity to prevent a stack overflow while retaining existing behaviour
* onShot method added to simplify creating blocks that should explode when shot at


70000hp:
HAH BOB,  TOLD YOU

PR #1483, opened 2024-05-10
Soot moment and bugs

70000hp:
**Fixes**

- Fixed smokestacks not working with machines with the new fluid pollution trait system
- Fixed crucible outputting ludicrous amounts of soot
- Fixed Industrial Combustion Engine outputting ten times more soot then intended (guh.)
- Fixed natural rampant scout spawning (for the billionth time) not working during the day
  -  Changed the base scout rate to account for that, it is recommended for people in previous rampant runs to change that base value

**Minor Changes**
 - Rampant now also has an increased pollution multiplier, and bombardiers spawns by default in it again
 - Switched turbofan to new fluid pollution system, mainly for consistency's sake, users may notice reduced pollution rates, enjoy!

### DIFF
--- a/src/main/java/com/hbm/config/MobConfig.java
+++ b/src/main/java/com/hbm/config/MobConfig.java
@@ -59,7 +59,7 @@ public class MobConfig {
 	public static boolean rampantMode = false;
 	public static boolean rampantNaturalScoutSpawn = false;
 	public static double rampantScoutSpawnThresh = 14;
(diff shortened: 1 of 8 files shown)


signed-firmware:
can we keep the 10x industrial combustion engine?

Martian64:
Does it also include the Scout Spawn Threshold being changed to its intended value, from 20.0 to 14.0?

PR #1485, opened 2024-05-11
fixed ricochet chance to work as intended

antonytvton:
Instead of using the odd hRic and lRic system, I made it such that it would take the percentage of the hit angle and ricochetAngle and use that in a weighted average to determine the bounce chance


antonytvton:
https://discord.com/channels/320301110759784449/708382585721716787/1238645590821437460
for the bug report

PR #1497, opened 2024-05-20
Hacker6400 patch 2

Hacker6400:
I' ve edited the new lang for the italian language but is not ready and i think it will take some time to be ready, but for now achievements, rbmk book, armors, battery info and detonator message is ready.
immagine_2024-05-20_162656156


Hacker6400:
If you see that I've eliminated somethings, it because I make some errors in the langs file

PR #1515, opened 2024-05-30
Reacher in 528

Ruslanllll:
Make reacher reduce radiation levels in 528 mode


SoB1nary:
it not working in 528 is intentional

RaidhoSketch:
Not making it reduce rads is honestly fair in 528, it's the pyrophorics thing that's unfair

Ruslanllll:
> it not working in 528 is intentional

In this case: Why is this code needed?
`			if(GeneralConfig.enable528 && reacher) {
				rad = (float) (rad / 49F);	//More realistic function for 528: x / distance^2
			}`

PR #1525, opened 2024-06-04
fenc

MellowArpeggiation:
Chainlink fences now automatically place posts at corners and junctions, otherwise rendering just the chain.
If you wish to manually place posts, just chuck the fence into a crafting table to convert it into a post, which will always render a post.
No new block IDs used, just uses block meta.


Vaern:
peak

gmhs022:
FINALLY!

HbmMods:
grafik

PR #1529, opened 2024-06-06
grate

MellowArpeggiation:
Sneak-click while placing a grate over a non-full block like a duct or cable to sink the grate flush with the floor!

Also works on ceilings


LePeeperSauvage:
I am going to kiss you

HbmMods:
grafik

PR #1531, opened 2024-06-07
ByteBuf string reader/writer + performance improvements

MellowArpeggiation:
over and over and over and over and over, like a monkey with a miniature cymbal


HbmMods:
image

mh yess

PR #1540, opened 2024-06-11
OC Compatibility part 2: the return of the `@Callback`

BallOfEnergy1:
So I'm going to leave this open for a bit until I can finish the radar compatibility fully, though artillery/turret compatibility is finished and tested.


HAJONK:
Very nice.
Will definitely add OC to my survival world

HbmMods:
Did you test this without OC installed? And for the turret control thing, does it "paralyze" active turrets when in use or does it rely on turrets being idle so they can be controlled?

BallOfEnergy1:
I did test without OC installed, though I will have to test the turret control tomorrow; sorry I didn't see the comment earlier.

BallOfEnergy1:
The turret control does stop turrets attacking when they are disabled using OC, though controlling targets of the turret is specific to the artillery turrets and does not change the target at that very moment. There is a critical bug in the radar compat though; those will be fixed in the next commit here.

BallOfEnergy1:
yerp ok that should fix it

HbmMods:
grafik

PR #1542, opened 2024-06-15
Fix powernet disconnections

MellowArpeggiation:
This PR fixes power net disconnections that occur when worlds other than the overworld are loaded (or when the overworld isn't loaded at all)


HbmMods:
image

PR #1576, opened 2024-07-10
Electrolyzer improvements part 2

70000hp:
Changes:

- Removed the dust byproduct from electrolysis, and slag byproducts only exist if neither main product can be melted (only non-metal bedrock ore currently)
- Added a recipe maker helper method, which bypasses the issue that lead to the dust byproducts existing in the first place


src/main/java/com/hbm/items/special/ItemBedrockOreNew.java
```diff
@@ -169,6 +169,11 @@ public static ItemStack extract(Object o, int amount) {
 			List<ItemStack> billets = OreDictionary.getOres(frame.billet(), false); if(!billets.isEmpty()) return fromList(billets, amount);
 			List<ItemStack> ingots = OreDictionary.getOres(frame.ingot(), false); if(!ingots.isEmpty()) return fromList(ingots, amount);
 		}
+		if(o instanceof ItemStack){
```
HbmMods, line 173: At the time when the enum is initiated, most items don't even exist yet, so using items or item stacks doesn't make a whole lot of sense.

70000hp: I only used that on my recipe maker, so it can handle any kind of item byproduct including crumbs

HbmMods: What do you mean? The function is supposed to convert `BedrockOreType`'s output objects into usable itemstacks, anything that falls outside of that doesn't belong there at all.

PR #1596, opened 2024-07-22
Fix electrolyser crashing due to a null output

Voxelstice:
should be self explanatory. i have no idea if this was done in-dev already


HbmMods:
That's already fixed in dev

PR #1610, opened 2024-08-03
New record 2

Hacker6400:
This pull request is for the ItemModRecord.java and introduce the new disc


HbmMods:
New music discs aren't planned, relatively few people use them and long music files inflate the file size by a noticeable amount. Besides, your PRs are missing item textures and the actual audio track, so even if I wanted to merge them it wouldn't work.

PR #1623, opened 2024-08-12
Broken emerald fragment fix

Creeper-banner:
The emerald bedrock ore fragment is broken and not considered a vaild bedrock ore fragment (e.g. unable to craft to emerald dust, red tint). This commit fixes this.


HbmMods:
Already resolved in dev

PR #1625, opened 2024-08-12
More Machine Configs!

RosaTryp:
Added machine configs for:
- Ashpit
- Industrial/Leviathan and regular Steam Turbine
- condenser, powered condenser and cooling towers

If a new condenser is added in the future that also extends TileEntityCondenser and no overrides for the IConfigurableMachine methods are added, there will be a second, useless "condenser" config option.


HbmMods:
grafik

RosaTryp:
Awwwwww cutiee

PR #1629, opened 2024-08-14
OC Flixes

BallOfEnergy1:
Changed the proxy system for OC Compatibility, now ports should be properly detected.
The above fixes the issue with orientations, massive component counts, and components being registered as null.
Fixed Refinery recipes button having the incorrect values.
Removed `Analyzable` implementation; it was a terrible idea.


BallOfEnergy1:
I am extraordinarily proud of the floppy disk system ❤️

src/main/java/com/hbm/handler/CompatHandler.java
```diff
@@ -43,6 +68,134 @@ public static FluidType intToSteamType(int arg) {
         }
     }
 
+    /**
+     * Allows for easy creation of read-only filesystems. Primarily for floppy disks.
+     * (Though maybe reading directly from VOTV drives as filesystems could be implemented. :3)
+     **/
+    private static class ReadOnlyFileSystem implements Callable<FileSystem> {
```
HbmMods, line 75: Does this work without OpenComputers installed? `FileSystem` is part of the class signature without an optional annotation, I'm not sure how it ends up behaving but it feels like it shouldn't work

MartinTheDragon: This *should* work for two reasons:

- Nested classes get compiled into separate class files. The `ReadOnlyFileSystem` class file doesn't ever get loaded if OC isn't there since it gets referenced in `FloppyDisk`, which is another nested class that gets compiled into a separate class file, which only gets loaded when `Loader.isModLoaded("OpenComputers") == true`
- Type erasure (generic type arguments don't get preserved and therefore aren't technically there at runtime)

*Insert no warranty and no liability disclaimer here*

HbmMods: good enough for me

PR #1658, opened 2024-08-31
Xenium Resonator radar Compat

9Valjew:
It was sitting half-finished on my Intellij for so fucking long so fuck it
- Xenium Resonator compatibility with the sat relay linker
- Takes the topmost block in the X and Z coordinates given, maybe one day I will make a way to select the Y level on radars with like a dial or something, probably not
- Also made the relay chip a variant of the satellite chip instead of its own item because there was no reason for it to be a separate item
- that's it I think? yeah?


signed-firmware:
its yummers

PR #1693, opened 2024-09-23
Fix #1690 (upgrading crates & mass storage)

### LINKED ISSUE #1690: Upgrading crates voids contents
It would be nice if upgrading steel crates to desh (and other similar recipes) kept the inventories intact. As of now, the inventory is reset silently

abel1502:
(Okay, apparently github doesn't recognize issue mentions in the title, so I'm tagging #1690 here again for the record)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,25 +1,28 @@
 # eclipse
-eclipse
-bin
-*.launch
-.settings
-.metadata
-.classpath
-.project
+/eclipse
+/bin
+/*.launch
+/.settings
+/.metadata
+/.classpath
+/.project
 
 # idea
-out
-*.ipr
-*.iws
-*.iml
-.idea
+/out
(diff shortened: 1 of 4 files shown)


HbmMods:
Does the crafting handler work in NEI? Because I'm pretty sure there's no support for custom recipes in the standard NEI handler.

abel1502:
It does, in my testing. Presumably because I inherit from `ShapedOreRecipe` and don't fundamentally change its workings either

abel1502:
That said, I was using the unofficial NEI (`NotEnoughItems-2.6.38-GTNH.jar`). I'll test it with the original as well, just to be sure

abel1502:
Yep, works with regular NEI too

HbmMods:
huh i didn't think it was that easy

Issue #1692, opened 2024-09-21
Conveyor grabbers miss every 5th item


abel1502:
I'd built a storage system using conveyor grabbers with specific items whitelisted, conveyors behind them and crates in front of them. However I observed a strange behaviour, where a few of the crates wouldn't get any items. I conducted a test in a creative world (screenshot below) and it confirmed that every 5 blocks along a conveyor line, the grabber's 20-tick delay would align in such a way that the item manages to pass straight by it with no interaction.

On the screenshot below, every grabber was set up with a whitelist allowing only the particular colored concrete variant placed atop it. No upgrades were used. Red scaffolds mark grabbers who didn't manage to grab their corresponding items.

![{3FCB8B16-15DB-49B6-83C8-A9E7ED2EE8F8}](https://github.com/user-attachments/assets/9f9b52bf-2fa6-4371-af45-7dff5cd4b1dc)

One solution could be to count the delay starting from the last tick an item was grabbed, rather than ticking every nth tick, as is the case now:

https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/fcf0f0d2d97d5837b8c315d7c95086d5261040fc/src/main/java/com/hbm/tileentity/network/TileEntityCraneGrabber.java#L66

Issue #1686, opened 2024-09-19
Builder's choice concrete's blast resistance not displayed


abel1502:
Unlike other blast resistant blocks, builder's choice concrete (`concrete_colored_ext`) doesn't have an orange line stating its blast resistance in its item description. This is due to it being registered without the `ItemBlockBlastInfo` class as its item form handler:

https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/ca14f47cbc7dbb1bdfa42ef0f68b4678eb122da5/src/main/java/com/hbm/blocks/ModBlocks.java#L2627-L2629

PR #1694, opened 2024-09-23
Cache tile entity network packets

abel1502:
While I'm at it, I might as well share the rest of my personal changes

When investigating high network lag on my server, I noticed a lot of machines constantly sent update packets every tick, even when nothing changed, and decided to optimize it. Now packets are only sent when something in them is different AND every 20th tick to accommodate for the fact some machines appear to be initialized incorrectly when unloaded and loaded by a client. (In my testing, filled fluid barrels did this, but I assume there might be more machines with this problem).

Also includes fixes for #1686 and #1692.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,25 +1,28 @@
 # eclipse
-eclipse
-bin
-*.launch
-.settings
-.metadata
-.classpath
-.project
+/eclipse
+/bin
+/*.launch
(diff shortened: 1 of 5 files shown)


BallOfEnergy1:
this will cause tens of merge conflicts when I try to update my fork ❤️

PR #1700, opened 2024-09-26
Revert all that gradle bullshit and do reflection instead

MellowArpeggiation:


HbmMods:
go go gadget: blindly deleting the master block in the merge conflicts and praying it works

PR #1710, opened 2024-09-30
Improve strand caster

abel1502:
I've found strand caster to be a useful machine, but quite inconvenient for automation due to its inability to cast less than 9 items at a time. I've added an upgrade that enables the caster to cast its buffer, even if it isn't full, once every 10 seconds. The crafting recipe and general balancing is up for debate, I kinda just chose things at random. I've playtested the changes, so they do work properly as is.


HbmMods:
Honestly that sounds like something that should just be the default option, no upgrade necessary.

abel1502:
That would be fine by me too. I assumed it was an intentional drawback to the strand caster -- you get better performance in bulk, but suffer when switching materials.

That said, now that I've drawn an upgrade and the gui textures, I kinda wish it would go into the mod)...

70000hp:
just do a lever or smth, having an upgrade just for that is a waste

abel1502:
Alright, I've removed everything related to the upgrade, now it's just a change to the baseline strand caster

HbmMods:
grafik

PR #1719, opened 2024-10-05
censorship fr*ckin rules ! ! !

Yoghurt4C:
# Valve screwed over the Artifact fanbase by cancelling the Artifact 2.0 beta without even allowing users to invite friends like they said they would in January 2021 and does not deserve your financial support for these gross consumer-unfriendly practices.
image
p.s.s : the bloody pad is a bonus for when the menstuation module is production ready (or spacefork gets merged upstream)


TheVitya2127:
Disgusting

BallOfEnergy1:
what??

PR #1720, opened 2024-10-05
Add new mode to conveyor grabbers

abel1502:
When using grabbers for storage automation, it is rather inconvenient to have to explicitly set up a whitelist for each, not to mention being limited to only storing 9 kinds of items per crate. I introduced a new mode to the grabber aimed at improving this: only grab items matching what's already stored in the connected inventory. Tested, works, as usual.

Also fixed the setting tool support for grabbers and extractors, now the whitelist-blacklist mode is copied along with the filter.


HbmMods:
What's the actual use-case for this? If I understand this right, then wouldn't that mean that if the items in the crate are used up faster than the grabber can provide them, the stacks would run out and the grabber would no longer grab anything?

Also I'm not sure if I understand this code correctly:
```
for(int i = 0; i < 9; i++) {
	ItemStack filter = slots[i];

	if(filter != null) {
		filterMode = matcher.modes[i];
		break;
	}
}
```
It grabs the mode off the first filter without actually using the filter? That feels like a bit of a weird way to implement this, and I'm not sure people would even be aware of this functionality at all. Besides that, it means that only one filter mode is applicable for the entire container which is a bit of a weird restriction given that the entire point is to provide more item filters than the standard 3x3 grid.

abel1502:
> What's the actual use-case for this? If I understand this right, then wouldn't that mean that if the items in the crate are used up faster than the grabber can provide them, the stacks would run out and the grabber would no longer grab anything?

The primary use case is storage, where only inputs are handled automatically and outputs are done manually by a player.

> Also I'm not sure if I understand this code correctly:
> 
> ```
> for(int i = 0; i < 9; i++) {
> 	ItemStack filter = slots[i];
> 
> 	if(filter != null) {
> 		filterMode = matcher.modes[i];
> 		break;
> 	}
> }
> ```
>
> It grabs the mode off the first filter without actually using the filter? That feels like a bit of a weird way to implement this, and I'm not sure people would even be aware of this functionality at all. Besides that, it means that only one filter mode is applicable for the entire container which is a bit of a weird restriction given that the entire point is to provide more item filters than the standard 3x3 grid.

I agree, this does feel quite weird. The idea was to allow to somehow switch between "item match" & "item & meta match" for the new grabber mode, but it ended up being unintuitive and, in my play-testing, not really useful. I'm all right with removing this functionality.

abel1502:
There, the weird bit's gone now

PR #1723, opened 2024-10-06
Add stair and slab variations to colored and builder's choice concrete

abel1502:
In retrospect, I see why you didn't bother to do this. If I knew how painful it'd be, maybe I would've tried this. But now it's done. It's a bit messy, but improving this part would require a much larger refactoring, and I'm not ready for that yet. As usual, I've tested it and it does work.

I also changed `BlockConcreteColored` to use `BlockEnumMulti` like `BlockConcreteColoredExt` does. The previous implementation basically did the same thing by hand.


abel1502:
Apparently, the original `BlockConcreteColored` numbered colors in reverse (so, meta n corresponded to color 15-n). I've now reordered the enum accordingly

HbmMods:
I consciously avoided making slabs and stairs for every single concrete variant because stairs are limited to one variant per ID (at most two with a custom implementation) and slabs to eight, which means that's a lot of extra block IDs we'd have to use. NTM already wastes a lot of IDs as-is with all material blocks still using the older way of doing variants, and I didn't want to make the problem worse. If you want stair variants for any other block, I recommend carpenter's blocks.

abel1502:
Fair point.

I think the `BlockConcreteColored` change should still be incorporated as an independent refactoring, though. As of now, it pointlessly duplicates the functionality of `BlockEnumMulti`.

abel1502:
I've come to think, would it be acceptable to put this in behind a config option (off by default)? This way, those with enough spare block ids (as is the case either when only using NTM, or with a mod like EndlessIDs) could enable colored stairs and slabs, yet it wouldn't cost anything to the average person. Since all recipes involving them are autogenerated, it should be pretty trivial. I'll probably reimplement it this way and make another PR, unless you have objections.

(Carpenter's block just feels too... frivolous? Having access to diagonal slopes and vertical slabs and whatnot detracts from the challenge of making minecraft structures for me)

PR #1724, opened 2024-10-08
Add support for all Blender f-curve functionality

MellowArpeggiation:
This adds the ability to author any arbitrary animation using the full set of Blender lerps, easings, etc!

Doesn't include any new animations, but the authoring process is the same as before.


HbmMods:
grafik

PR #1736, opened 2024-10-17
Custom mats

FOlkvangrField:
Attempted to add a system that allows players to customize materials, with material forms only supporting items generated using the ItemAutogen class


HbmMods:
maybe some day, but not like that

PR #1751, opened 2024-10-31
Allow collecting Euphy's statue with a red key

abel1502:
I've just found my first `elb_statue` in a meteor dungeon in a survival world, and I felt a bit upset that I had no way to bring it home with me (from another planet, as I'm playing on JamesH2's fork). So I thought of a balanced way to enable this: now right-clicking the statue with a red key (cracked or regular, and cracked ones are consumed in the process) drops the statue (the bare variant), as well as whatever it came equipped with. If you'd like, I could change it to drop the statue with the items equipped instead. I thought this way is a bit more interesting, since now the player gets the option to take a little shortcut in crafting euphemium armor, if they were lucky enough to find a statue in a meteor dungeon. (The plates are still locked behind osmiridium, but no yharonite is needed for the watch).

### DIFF
--- a/src/main/java/com/hbm/blocks/generic/DecoBlockAlt.java
+++ b/src/main/java/com/hbm/blocks/generic/DecoBlockAlt.java
@@ -1,5 +1,6 @@
 package com.hbm.blocks.generic;
 
+import java.util.ArrayList;
 import java.util.Random;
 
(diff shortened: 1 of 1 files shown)


HbmMods:
sounds fair

PR #1752, opened 2024-11-01
Fix some typos

abel1502:
I noticed the broken watch said "It's glass has ..." instead of "Its", and I had nothing better to do than correct this in the entire project.


DoubleHypercube:
...these mean the same thing. " it's " is a contraction of "it is"

PewPewCricket:
Mental retardation

abel1502:
@AutumnalModding yes, and "its" is "belonging to it" (for example, "look at its wings"), which is a different thing, but is often confused with "it's"="it is" due to identical pronunciation

PR #1753, opened 2024-11-02
Jetpack-like wings

abel1502:
Now limp and murky wings can not only be worn, but also attached to armor. Functionally, they now share a base class with jetpacks (`JetpackBase`, except everything related to fuel has been extracted into `JetpackFueledBase`, which the actual jetpacks now inherit from).

This means they no longer act as defensive armor when worn on their own and don't get damaged when the player is attacked, but to me this makes more sense than otherwise. The previous behavior can be attained by attaching them to actual armor.

As per usual, I've tested this and it works perfectly, both when worn standalone or attached to a chestpiece.


WolfEclipses:
This feels unbalanced, like, being able to attach them to armor ruins the balancing imo.

HbmMods:
oh hell yeah this is what i wanted to do ages ago

Vaern:
> This feels unbalanced, like, being able to attach them to armor ruins the balancing imo.

nah this just means you have more choice for the jetpack slot and you can use wings with armor, which is awesome

WolfEclipses:
> > This feels unbalanced, like, being able to attach them to armor ruins the balancing imo.
> 
> nah this just means you have more choice for the jetpack slot and you can use wings with armor, which is awesome

That's fair

PR #1767, opened 2024-11-14
Custom fluid traits fix

FOlkvangrField:
Fixed some custom fluid traits buds, like the JSON output error of the polluting trait and the java.lang.InstantiationException issue of the PWRModerator and Pheromone traits. Mean while, the hot reload function of hbmFluidTypes.json and hbmFluidTraits.json has been added.


FOlkvangrField:
@HbmMods I noticed that the latest release log does not seem to mention the custom fluid fix and the addition of custom fluid hot reload functionality. Do these contents need to be improved and then published in the future?

PR #1770, opened 2024-11-16
hangman model + texture

Vaern:
CoC:
- chudman mogel + fanum taxture


HbmMods:
please never say any of these words ever again

PR #1806, opened 2024-12-16
New mini nuke textures

TheVitya2127:
- Added new mini nuke textures
*It looks something like this:*
IMG_20241206_120203


HbmMods:
grafik

TheVitya2127:
By the way, you are right. The textures are terrible.

PR #1838, opened 2024-12-28
Add IConfigurableMachine to multiple machines

Agarmal:
Well, as the title says, I added `IConfigurableMachine` to 28 machines in total (the 29th one is just a comment on `ArcWelder` saying that I couldn't add `IConfigurableMachine` to it bcs I would have to change way too many lines of code).

It says that a lot of lines have been changed, but that's not actually the case bcs I had upload the modified files entirely instead of just the modified part cause I'm an idiot.

I've built and tested the code, and as expected, it runs with no problems... tho, if there's any issue just tell me to fix it before merging or smth...

### DIFF
--- a/src/main/java/com/hbm/tileentity/machine/TileEntityMachineAmgen.java
+++ b/src/main/java/com/hbm/tileentity/machine/TileEntityMachineAmgen.java
@@ -1,6 +1,9 @@
 package com.hbm.tileentity.machine;
 
+import com.google.gson.JsonObject;
+import com.google.gson.stream.JsonWriter;
 import com.hbm.blocks.ModBlocks;
+import com.hbm.tileentity.IConfigurableMachine;
 import com.hbm.tileentity.TileEntityLoadedBase;
 import com.hbm.util.CompatEnergyControl;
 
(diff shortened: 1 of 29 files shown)


BallOfEnergy1 (commented):
Most of the code looks fine just from looking over it, though at times it does seem a bit excessive. Having a config for max power on something like `TileEntityMachineAmgen` seems a bit useless. I'll look through the code a bit more thoroughly so bob doesn't have to deal with the diff structure, though IDK if it helps at all.

Agarmal:
I mean... I just went through a few machine files and turned static finals and some other variables into configurables, plus I did not change too much code to keep it simple. The structure is almost entirely the same, except for those values.
And also, it's true that I made it for some useless ones like `TileEntityMachineAmgen`, and it's because I did not do too many checks to see if they're useful at all since it's a pretty fast thing to write.

I just wanted to contribute, and I saw an opportunity with this.
image

I hope to not be causing too much trouble 😅

Agarmal:
I can still return files like `TileEntityMachineAmgen` to their original state in case it's necessary.

HbmMods:
Looks fine at a quick glance, just have to put that on the back burner until we have #1733 sorted out.

Agarmal:
Closed this PR because I realized it would cause issues with newer commits... I'll maybe do it again in the future.

PR #1845, opened 2025-01-02
Fix effect conflicts with Night Vision Goggles in the M1TTY Environment Suit

Agarmal:
So... the thing is that if you put a NVG on a M1TTY EnvSuit, it does not work because the M1TTY's water night vision overrides the NVG's effect. It was already getting on my nerves because I couldn't just slap NVGs in it, and I made this fix.


Agarmal:
Btw, I added a comment on the ArmorModHandler file... it shouldn't be an issue tho.

PR #1859, opened 2025-01-10
Update en_US.lang

Hacker6400:
before
i added the 3 fire key in the en_US.lang


SmiteX134:
I would've called it Primary fire, Secondary fire and Aim

PR #1871, opened 2025-01-15
Fix RBMK Crane Console

FOlkvangrField:
Fixed RBMK Crane Console moving problem. I actually don't know how the new packet sending code works. I compared the posLeft variable with the similar posFront variable and found that it performed an additional operation during deserialization, so I removed it, and this fix the problem that RBMK Crane Console can't move left or right.


BallOfEnergy1 (approved):
This does seem to fix an oversight when deserializing the data that could cause a buffer underflow, looks good to me.

PR #1873, opened 2025-01-16
Add vent carbon dioxide function in Zirnox for OpenComputers

Agarmal:
yes,
took me like 5 minutes (too much work ik).
image

The function is called `ventCarbonDioxide`.

### DIFF
--- a/src/main/java/com/hbm/tileentity/machine/TileEntityReactorZirnox.java
+++ b/src/main/java/com/hbm/tileentity/machine/TileEntityReactorZirnox.java
@@ -516,11 +516,9 @@ public class TileEntityReactorZirnox extends TileEntityMachineBase implements IC
 	@Callback(direct = true)
 	@Optional.Method(modid = "OpenComputers")
 	public Object[] ventCarbonDioxide(Context context, Arguments args) {
-		int ventAmount = args.optInteger(0, 1000); // Get how much CO2 to vent in mB (1000mB default)
+		int ventAmount = Math.max(Math.min(args.optInteger(0, 1000), carbonDioxide.getMaxFill()), 0); // Get how much CO2 to vent in mB (1000mB default), clamp between 0 and carbonDioxide's max fill.
 		int fill = this.carbonDioxide.getFill();
-		this.carbonDioxide.setFill(fill - ventAmount);
-		if(this.carbonDioxide.getFill() < 0)
-			this.carbonDioxide.setFill(0);
+		this.carbonDioxide.setFill(Math.max(fill - ventAmount, 0)); // Make sure it isn't a negative number.
 		return new Object[] {};
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/hbm/tileentity/machine/TileEntityReactorZirnox.java
```diff
@@ -513,6 +513,17 @@ public Object[] setActive(Context context, Arguments args) {
 		return new Object[] {};
 	}
 
+	@Callback(direct = true)
+	@Optional.Method(modid = "OpenComputers")
+	public Object[] ventCarbonDioxide(Context context, Arguments args) {
+		int ventAmount = args.optInteger(0, 1000); // Get how much CO2 to vent in mB (1000mB default)
```
BallOfEnergy1: I advise setting a limit to how high of an integer can be inserted, wouldn't want an overflow.

src/main/java/com/hbm/tileentity/machine/TileEntityReactorZirnox.java
```diff
@@ -513,6 +513,17 @@ public Object[] setActive(Context context, Arguments args) {
...
+		int fill = this.carbonDioxide.getFill();
+		this.carbonDioxide.setFill(fill - ventAmount);
```
BallOfEnergy1: This should also be changed to not set the fluid tank to a negative value before correcting it, maybe you could add an intermediate variable in here.

src/main/java/com/hbm/tileentity/machine/TileEntityReactorZirnox.java
BallOfEnergy1: Also, negative numbers need to be clamped away!!!

src/main/java/com/hbm/tileentity/machine/TileEntityReactorZirnox.java
```diff
@@ -513,6 +513,17 @@ public Object[] setActive(Context context, Arguments args) {
 		return new Object[] {};
 	}
 
+	@Callback(direct = true)
```
BallOfEnergy1, line 517: Normally I don't recommend using fully direct callbacks for functions that directly modify the TE data, as it can lead to lag issues. You can leave it as is, though I suggest adding a limit of 4 or so using the following code.
`@Callback(direct = true, limit = 4)`

BallOfEnergy1 (changes requested):
Other than the changes requested, this seems to be in working order. Hope to see it in the mod soon.

src/main/java/com/hbm/tileentity/machine/TileEntityReactorZirnox.java
Agarmal: Oh uhm.. Sorry, I'll fix that ASAP..

Agarmal:
> Other than the changes requested, this seems to be in working order. Hope to see it in the mod soon.

Changes commited, just tell me if there are other issues. (**BTW**, the control function used by the gas vent valve in the GUI corrects the `fill` value after setting it, which is the reason mine also was like that)

Agarmal:
I used `Math.min(...)` and `Math.max(...)`, but I can use conditionals if it gets too complicated.

HbmMods:
You can use `MathHelper.clamp_int(value, min, max)` for stuff like that

Agarmal:
> You can use `MathHelper.clamp_int(value, min, max)` for stuff like that

Oh, ty for the tip

PR #1891, opened 2025-01-25
Add OpenComputers Compatibility for Combustion Engine 2

t3435ryt:
Adds ten new OpenComputers methods for the combustion engine.
* Now without infinite fuel bug (thank you for your suggestion gamma)


t3435ryt:
fail again: I was stupid and sent it before building

PR #1894, opened 2025-01-26
Fix foundry channel clogs

MellowArpeggiation:
This PR prevents foundry channels from accepting new pours from a crucible/electrolyser/etc until the whole channel has completely emptied the current material, meaning that as long as there are basins/strandcasters can accept new material, clogs will not occur.


MellowArpeggiation:
it might be necessary for players to break and replace existing channels due to the NBT loading `Stone` as the neighbor type, causing nothing to flow initially.

edit: not anymore baybeee

HbmMods:
grafik

PR #1917, opened 2025-02-03
Simple Github action to build on push & pr

TheCrazyInsanity:
You only get 2k minutes a month without paying, but you don't seem to be using actions anywhere else so might as well right?


HbmMods:
gradle scares me

grafik

HbmMods:
is there a way to do this maybe only once a day because right now when i handle PRs i usually get like three consecutive runs on the action

PR #1924, opened 2025-02-06
PA Detector fixes and improved diagnostics

Dash1269:
Fixed a bug where some code paths in `TileEntityPADetector.onEnter` would not call `particle.crash` causing the diagnostics in the PA source to not update; New diagnostics/PAStates were also added for these code paths.

Fixed the PA detector not consuming power on a successful recipe.

Slightly refactored to make the PA detector code more readable.

I have tested these changes in game and they seem to work fine :)


HbmMods:
grafik

i already fixed that like 2 hours ago but honestly i like your solution better

PR #1936, opened 2025-02-11
Rotary Furnace Fuel Tweak

70000hp:
**Changes**

Certain solid fuels (solid, rocket, balefire) speed up the rotary furnace, but require more steam 
Steam requirements scale with speed nonlinearly, the function being 13 log(fuel speed) + 1
-  This makes parallelizing the furnace more efficient on steam consumption, which can become high in large setups
-  This specific function makes solid fuel the most steam inefficient fuel, and the better the fuel, the less steep the difference  between speed and consumption, with balefire rocket fuel offering the best ratio out of the three


HbmMods:
fiiiine

HbmMods:
when undraft

70000hp:
i may be stupid

HbmMods:
why did you do that

PR #1943, opened 2025-02-17
added mas-36 model, texture

Vaern:
CoC:
- added model and texture for the mas-36 cr39 carbine
  - woooooo sexy dithering and standard-compliant texel density


HbmMods:
grafik

PR #1959, opened 2025-02-27
spent casings prrrrrr!!!

Vaern:
CoC:
 - Gave ParticleSpentCasings rotational velocity and randomized bounces
   - Uses a slimmed version of MoveEntity, with special code handling bounces, loss of rotvel, etc.
   - messed with setSize to prevent casings from hovering a great distance - seems to work fine even with the larger shells
   - setBounceMotion now serves as a multiplier for the bounces, meaning e.g. artillery shells won't spin nearly as fast as handgun casings
 - Gave CasingCreator and CasingEjector initial rotational velocities
   - added it as an argument for the former, plus all of them set in the orchestras that hopefully make sense
   - added as default for turrets cuz CasingEjector kinda sucks
 - the most important change of all: muzzle/gap flashes now have fullbright. ur all welcome


Vaern:
oh also set casing models as vbos!!!

HbmMods:
grafik

PR #1960, opened 2025-02-27
rotary furnace fuel thing

70000hp:
same pr, not borked


BallOfEnergy1:
cmon buddy youll get it one of these times, just keep opening more prs

PR #1961, opened 2025-02-28
Performance boost for RBMKs (thanks DEKO)

MellowArpeggiation:
In making the RBMK neutron system more robust in cases of reactors in many dimensions, more hashmap lookups than necessary were added. This PR should cut it down to size by passing by reference instead where appropriate.

Performance improvement: 68ms over 60 seconds -> 8ms over 60 seconds


HbmMods:
grafik

PR #1965, opened 2025-03-03
Redstone-over-Radio OpenComputers integration + Mining satellite NEI handler

Toshayo:
This PR improves compatibility of Redstone-over-Radio Receiver/Transmitter with OpenComputers. It also adds a NEI handler that shows mining satellite production with chances for each ore.

image


HbmMods:
What does it do when there's more than 18 items in the satellite item pool?

Toshayo:
When there are more than 18 items (as it is for the normal miner satellite) the 18th-nth items will share slots with 18 first ones (cycling)

PR #1969, opened 2025-03-04
Make HazardRegistry factory methods public

iocmet:


iocmet:
@HbmMods Explain why you don't want to merge if you don't

burningwater202:
you only made the pull relax 9 hours ago chill

iocmet:
Ikr, it seemed like developer was active though

HbmMods:
Could you at least explain what you're trying to do with your PR

iocmet:
Make these methods public so other mods can use them?

HbmMods:
* The methods you made public are just minor utility for shoving data into several registry maps, most of which are already public
* You can just skadoodle around visibility modifiers using reflection
* All the methods for adding recipes aren't API, they're only really intended for internal use, and may change at any time (in the case of the chemical plant, probably soon). NTM doesn't really have a robust system for that, since addons aren't really all that common
* Depending on the load order, recipes either break easily or never work at all
  * If the recipes are added before `SerializableRecipe.initialize()`, the init process will write over all the recipes, since the process involves clearing the recipe list
  * If the recipes are added after, they will work, however they won't be added to the recipe template file, and as soon as `/ntmreload` is used, the recipes are once again deleted

A much more reasonable approach would look as follows:
* A registerable recipe creator object which would be called by `SerializableRecipe.initialize()` during the actual process, ensuring the recipes are added and not overwritten, used in the template file, and re-created on recipe reload
* Recipes would be created via a separate class that is part of the API with a never-changing method signature to prevent any issues arising from the recipe format or machines changing. The recipe creator just has to use the API, and the API would take all the data that's thrown at it and make the best of it.

iocmet:
1. I know
2. Yes but this is not intended way of adding recipes and NTM is actively developed mod, so i thought it is better to change that
3. Ok, i did not know that and used them as API
4. I know
Okay

iocmet:
> * A registerable recipe creator object which would be called by `SerializableRecipe.initialize()` during the actual process, ensuring the recipes are added and not overwritten, used in the template file, and re-created on recipe reload
> * Recipes would be created via a separate class that is part of the API with a never-changing method signature to prevent any issues arising from the recipe format or machines changing. The recipe creator just has to use the API, and the API would take all the data that's thrown at it and make the best of it.

What about dynamic recipes registration after initialization?

BallOfEnergy1:
my email getting blown tf up from this thread

iocmet:
Imagine using github web from phone

HbmMods:
Why would you need to change recipes after registration?

iocmet:
I think there can be some use cases besides crafttweaker?

HbmMods:
Why would anyone use Craft Tweaker, we got our own configs

iocmet:
That's the reason i said there can be use cases *besides* crafttweaker, also vanilla recipes can be dynamically registered

PR #1970, opened 2025-03-05
Make RBMK Rods actually be a SimpleComponent

kelllllen:
I was wondering why I couldn't connect OC cables to a rbmk rod, turns out the tile entity didn't implement SimpleComponent. Tested, works.


MellowArpeggiation:
my bad, my IDE marked it as a redundant interface since it is included in CompatHandler.OCComponent so I removed it when fixing RBMK unloading issues (must be OC doing something weird when inferring type inheritance)

PR #1973, opened 2025-03-07
Fix comically long 9mm reload animation snapping

MellowArpeggiation:
this should be the _last_ bezier fix.

I'm gonna write a wiki page on how to author these animations now byeee


HbmMods:
grafik

PR #1981, opened 2025-03-12
There i fixed chunkloading drones

kelllllen:
The code can probably be extended to other stuff like missiles and artillery, but I don't even know if those are broken right now, so I'm not gonna touch them.
Added a util class that gets all the chunks along a line (the current pos + motion) with an amount of padding so the drones or whatever don't get stuck on corners. A padding of 4 blocks seems to work pretty well for drones, and I have had it get stuck with 3, so 4 it is for now.
Also help my ide is doing some weird shit with indentations.


HbmMods:
I don't really see how that would fix it since "moving" an entity just means changing its position once within a tick, and as long as the chunk at that position is loaded, the entity also remains loaded. Missiles and artillery rockets use the same principle, and they work just fine. However I've tried fixing these fucking drones like three times already and they never once worked so I might as well

PR #2000, opened 2025-03-21
Leather Flavored Handlers: Blue Flavor

Axtrine:
Just a general fix for how the RBMK guide worked, nothing too big really.

It’s more neat now rather than before

### DIFF
--- a/src/main/java/com/hbm/items/tool/ItemGuideBook.java
+++ b/src/main/java/com/hbm/items/tool/ItemGuideBook.java
@@ -87,60 +87,30 @@ public class ItemGuideBook extends Item implements IGUIProvider {
 		
 		return pages;
 	}
-	
+
 	//TODO: Make sure this is all correct
 	public static List<GuidePage> statFacRBMK() {
-		
-		List<GuidePage> pages = new ArrayList();
-		pages.add(new GuidePage().addTitle("book.rbmk.title1", 0x800000, 1F)
-				.addText("book.rbmk.page1", 2F)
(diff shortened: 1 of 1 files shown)


Axtrine:
I'll do the rest of the books once I can use my computer again, providing I remember :p

Axtrine:
On second thought, if you consider merging it wait until I finish the handler for the other books.

HbmMods:
In the original definition, page 13 had no image, and page 14 has image 13, with image 14 being absent entirely. In the new one, page 13 has image 13, and page 14 doesn't even exist.

Besides, what does this PR even do? It just looks like you're replacing the builder-style setup for the pages with a fixed parameter list, which is exactly what I tried to avoid.

Axtrine:
I think I'm gonna close this until I fix it anyway.

Axtrine:
> In the original definition, page 13 had no image, and page 14 has image 13, with image 14 being absent entirely. In the new one, page 13 has image 13, and page 14 doesn't even exist.
> 
> 
> 
> Besides, what does this PR even do? It just looks like you're replacing the builder-style setup for the pages with a fixed parameter list, which is exactly what I tried to avoid.

Oh sorry, it was just an idea that was passed to me so I thought I'd try to help.

Axtrine:
> In the original definition, page 13 had no image, and page 14 has image 13, with image 14 being absent entirely. In the new one, page 13 has image 13, and page 14 doesn't even exist.
> 
> 
> 
> Besides, what does this PR even do? It just looks like you're replacing the builder-style setup for the pages with a fixed parameter list, which is exactly what I tried to avoid.

Oh sorry, it was just an idea that was passed to me so I thought I'd try to help.

PR #2001, opened 2025-03-21
G3 Attachments

Vaern:
CoC:
- Actually added the g3 attachments sprite and its respective attachments
  - Noticed that the uv was fucked in-game, re-exported the model with fixed model names
    - The bolt actually renders now!
  - Adjusted in-game icon to fit the big-ass silencer
 - Fixed/changed animations
   - Animation of the bolt separated from the charging handle's plug, meaning reload and dryfire now properly uses the charging handle
     - In addition, firing actually shows the bolt slam back and forth


HbmMods:
grafik

PR #2005, opened 2025-03-22
Toolboxes!

BallOfEnergy1:
BOO! did i spook ya?

Completely revamped toolboxes, now they act as portable hotbar switchers.

Not entirely finished, I plan to change the GUI tomorrow, but as of now, everything was coded *in two days*.


BallOfEnergy1:
You'll have to review this before merging, but as far as I see it should be finished.

PR #2010, opened 2025-03-24
Fix fluidnet bugs

MellowArpeggiation:
Fixes:
* Fluid valves not deactivating (was removing a non-existent power node instead of the fluid node)
* Compatibility shim for single blocks was not subscribing to the blocks _around_ itself, breaking single block turbines and condensers
* Fluid tanks next to each other in buffer mode eating fluids

Also changes some behaviours to make them more intuitive and give the player more control over power/fluid propagation:
* Energy storage blocks won't propagate power through themselves if not set to buffer mode
   * This means power won't skip over storage blocks that are set to not be buffers
   * Players can circumvent this by placing cable blocks _around_ storage blocks to skip them, which should be more intuitive, and means the blocks can be used as rudimentary switches as well if desired

### DIFF
--- a/src/main/java/com/hbm/tileentity/network/TileEntityFluidValve.java
+++ b/src/main/java/com/hbm/tileentity/network/TileEntityFluidValve.java
@@ -1,26 +1,27 @@
 package com.hbm.tileentity.network;
 
-import api.hbm.energymk2.Nodespace;
+import com.hbm.uninos.UniNodespace;
+
 import net.minecraft.block.Block;
 import net.minecraft.world.World;
 
 public class TileEntityFluidValve extends TileEntityPipeBaseNT {
-	
+
 	@Override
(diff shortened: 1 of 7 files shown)


src/main/java/com/hbm/tileentity/network/TileEntityFluidValve.java
```diff
@@ -1,26 +1,27 @@
...
 	}
 
 	public void updateState() {
-		
+
 		this.blockMetadata = -1; // delete cache
-		
+
 		if(this.getBlockMetadata() == 0 && this.node != null) {
-			Nodespace.destroyNode(worldObj, xCoord, yCoord, zCoord);
```
HbmMods, line 19: the fuck was i doing here

HbmMods:
grafik

PR #2012, opened 2025-03-24
More NBTStructure improvements from OUTER SPACE

MellowArpeggiation:
A whole mixed bag of stuff and things:
* NBTStructures now support selection/placement priority properly
* Perfectly abutting pieces will not generate a fallback piece, meaning grid layout jigsaws now work!
* NBTStructure jigsaw debug build all pieces method
* when using the construction wand with structure air blocks, it will only replace air, making it much easier to fill structures quickly
* a whole bunch more blocks support rotations properly, including:
    * tripwires
    * vanilla-style mod doors
    * RBMK parts (in that they _don't_ rotate)
    * electric press
    * macklewaves
    * plushies! (also they randomise too now)
* you can now repair generated industrial lights by interacting with them, instead of having to break and replace them
* fixed selection priority order being backwards (1 is now picked before 0, as intended)
* fix potential crash caused by caustic ass mods fucking with TEs while loot is attempting to generate

### DIFF
--- a/src/main/java/com/hbm/blocks/generic/BlockModDoor.java
+++ b/src/main/java/com/hbm/blocks/generic/BlockModDoor.java
@@ -4,6 +4,7 @@ import java.util.Random;
 
 import com.hbm.blocks.ModBlocks;
 import com.hbm.items.ModItems;
+import com.hbm.world.gen.INBTTransformable;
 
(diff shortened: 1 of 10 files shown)


Lazzzycatwastaken:
The last thing actually happened to me so gj fixing it 
image

PR #2014, opened 2025-03-25
Fixed crate item movement prevention applying to all held items

Toshayo:
Fixes an issue where when you open a placed crate you cannot move the item in your hand. This PR allows movement of non crate items but will still have the issue of not allowing to move a crate held in hand while opening a placed crate.


BallOfEnergy1:
oops, my bad

PR #2018, opened 2025-03-27
RBMK console fuel rod xenon bar

Voxelstice:
makes the xenon bar on fuel rods work in the RBMK console
javaw_5wyqifnuoE

or atleast that's what i think it is. correct me if this bar is for something else


BallOfEnergy1:
wait what I've never seen this before
it's been broken for all this time
incredible

Voxelstice:
it's actually never been coded in. i only noticed this because i've had to dig into the texture for a thing i was doing

HbmMods:
ough

PR #2021, opened 2025-03-28
Full RBMK crafting chains

MellowArpeggiation:
this PR adds 2 new recipe handlers: Nuclear Waste Decay and RBMK Rod Disassembly, and updates the Spent Fuel Pool Drum recipe handler make a full end-to-end recipe chain for RBMK rods. Features include:
* only showing cooling recipe if the rod is too hot to uncraft
* and vice versa, only show uncrafting recipe if the rod is cool enough
* shows all possible depletion levels, with minimums (eg: the fully depleted pellets are shown to come from 80%+ depletion, and xenon above 50% becomes _High Xenon Poison_)

no more will new players have to wiki it up for finding their first bismuth nugget (well I mean they probably still will because it's a long ass chain but hey)


HbmMods:
grafik

PR #2037, opened 2025-04-01
Fleija remodel

legendarydoge30:
its a fleija remodel


iocmet:
It has renderer registered in ItemRenderLibrary

iocmet:
https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/master/src/main/java/com/hbm/render/item/ItemRenderLibrary.java#L389

legendarydoge30:
oh i didnt see that
ig that could just be removed though?

iocmet:
Yes ig?

PR #2098, opened 2025-04-21
PA Dipole tweaks and particle serialisation

Dash1269:
Fixes the particle source not saving the current particle on unload, causing inflight recipes to be deleted.

Changes the behavior of dipole magnets to only apply the coil power penalties when it's redirecting the particle.

This change allows for single loop particle accelerators without having to meet the minimum length requirement while injecting into the loop from the particle source, as long as it's inline with the direction of the loop.

I've seen multiple people confused as to why their particle was immediately crashing upon leaving the particle source, and this should fix that issue.
It also allows slightly more convenient particle routing for more complex accelerator setups :)

Let me know if you need any further clarification or changes made.


70000hp (approved):

DrIshirozaki:
This explains a lot

PR #2099, opened 2025-04-21
fix for Ticking entity crash when spawning skeletors with guns

Lazzzycatwastaken:
https://github.com/user-attachments/assets/0b824faf-aeac-4aba-ae66-0addadecdf50
also increased the chances of gun skeletons spawning because when i tested it just now with like 50 soot it spawned legit none

### DIFF
--- a/src/main/java/com/hbm/main/ModEventHandler.java
+++ b/src/main/java/com/hbm/main/ModEventHandler.java
@@ -417,7 +417,7 @@ public class ModEventHandler {
 			}));
 
 			ItemStack bowReplacement = getSkelegun(soot, world.rand);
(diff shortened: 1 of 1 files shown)


DrIshirozaki:
Every once in a while my refinery skyrockets to like 60 soot and the skeletons become weaker because the shitty guns they wield are worse than arrows
Screenshot_20240114_195528_One_UI_Home

MellowArpeggiation:
it's not the guns, it's the inbuilt inaccuracy debuff to stop them being annoying as fuck, this can be adjusted

MellowArpeggiation:
> also increased the chances of gun skeletons spawning

the odds look unchanged to me?

PR #2127, opened 2025-05-02
Sliding blast door rail support

Toshayo:
This PR allows rails to be placed on the Sliding blast door base. When the door closes, it breaks the rail, so it's still blast proof. When the door reopens, the rail is placed back (preserved rail block name and metadata).

Feature testing:
- Singleplayer
  - Normal, special (powered, activator) and HBM's rails (high speed, wooden, etc) removed and placed
  - Rail rotation preserved
- Multiplayer
  - Same tests, everything seems to work

The PR also includes the small fix of a tileentity not being saved, as this fix was already implemented. Feel free to tell if it's already on your local repository so i can remove it to avoid merging issues.


HbmMods:
nah

PR #2128, opened 2025-05-03
color change to scumfold

Lazzzycatwastaken:
OLD:
image

NEW:
image

why? 1.need darker colors for the desert atom rework 2. no one likes these ultrabright crayon ass colors on these things, they dont fit into most builds


HbmMods:
i think it kinda looks not very good

also i'd like to be involved in the structure rework because unlike most of the other structures that one's actually pretty important

Lazzzycatwastaken:
can i add you to the gc where were doing the structure? Or do i just relay info somewhere elde

HbmMods:
either is good so long as i can see what yall are doing and have input

PR #2141, opened 2025-05-09
Prevents the automatic buzz saw from cutting saplings

CGPoly:
The automatic buzz saw can't be used to create automated tree farms because it detects and cuts down saplings. Adding a case for saplings to the shouldIgnore function fixes this behaviour.


Vaern:
I've made automated tree farms with it, you just need to keep the blade a block above the saplings...

HbmMods:
I find it honestly confusing how you managed to figure out how to set up a workspace, edit the mod and make that change, but not the fact that you can just place the buzzsaw one block higher so it doesn't annihilate all the saplings.

PR #2152, opened 2025-05-18
Tool abilities GUI

abel1502:
A rework of how tool abilities work, staying faithful to the original but adding new useful degrees of freedom.

- Tools with abilities now allow customizable presets.
- Tool abilities are split into area abilities (stuff that determines what blocks are mined) and harvest abilities (stuff that affects the drops of the blocks mined). Area and harvest abilities can be used alongside each other (like, for example, veinminer + silk touch).
- By default, the presets are the same as they were before (one for off, and one for each individual abilities), except their order is now made consistent across all tools (area abilities first, then harvest abilities -- though it can be changed if necessary).
- Right click and shift-click with a tool retain their functionality (next preset and first preset). Alt-click now brings up the GUI.
- Up to 99 presets are allowed per tool.
- The GUI easily accommodates new abilities (add an icon to the texture, add an entry to a list of buttons with its coordinates, and voilà), the screen automatically scales horizontally to fit everything automatically.
- The GUI is intuitive thanks to tooltips for all abilities and buttons.
- Scrollwheel is supported to cycle through presets more quickly.
- The GUI has nice interactive highlights and sounds.
- I mean just look at it, isn't it gorgeous!
- On the architectural side, abilities are now effectively represented by an enum and an integer level. (They are in fact static final objects and not enums, but the point is, no new object is created for an ability). A level has a customizable mapping to the ability's effect. For currently existing abilities, the maximum level in use is 7 (well, 6, but 0-based). The GUI allows up to 10 levels, but more can be added easily.
- Weapon abilities are also remodeled this way for consistency, but since they're always active, they have no GUI to them.
- Both the GUI and the logic are robust. Nothing breaks on out-of-bounds values, and attempts to cheat with abilities are caught and rectified server-side.
- Playtested to the best of my abilities (Which is less than it might suggest, so feel free to report anything I missed, and I'll fix it).

Here's a little gallery. (Sorry it's slightly wonky, but I can't be bothered to crop images by hand)

screenshot

screenshot

screenshot


gradle.properties
```diff
@@ -1,67 +1,67 @@
...
+credits=HbMinecraft,\
+  \ rodolphito (explosion algorithms),\
+  \ grangerave (explosion algorithms),\
+  \ Hoboy (textures, models),\
+  \ Drillgon200 (effects, models, porting),\
+  \ MartinTheDragon (calculator, chunk-based fallout, bendable cranes, pipe improvements, PWR sounds),\
+  \ Alcater (GUI textures, porting),\
+  \ MellowArpeggiation (new animation system, turbine sounds, sound fixes, industrial lights, better particle diodes),\
+  \ Pheo (textures, various machines, models, weapons),\
+  \ V�r (gas centrifuges, better worldgen, ZIRNOX, CP-1 parts, starter guide, new cyclotron, weapon animations),\
```
Vaern: to prevent stuff like this, you'll want to change your IDE's encoding to use utf-8 instead of cp1252 :)

src/main/java/com/hbm/items/tool/ItemToolAbility.java
```diff
@@ -1,51 +1,74 @@
...
 	protected boolean isShears = false;
 	protected EnumToolType toolType;
 	protected EnumRarity rarity = EnumRarity.common;
 	//was there a reason for this to be private?
 	protected float damage;
 	protected double movement;
-	protected List<ToolAbility> breakAbility = new ArrayList() {{ add(null); }};
-	protected List<WeaponAbility> hitAbility = new ArrayList();
-	
+	protected AvailableAbilities availableAbilities = new AvailableAbilities().addToolAbilities();
```
Vaern, line 69: Initializing `availableAbilities`, even though it's only the 'none' ones, means that `!toolAbilities.isEmpty()` will never be true for tools without any abilities. As such, tools like the steel pickaxes have tooltips mentioning the abilities despite having none, and similarly allow you to open the gui.

This is probably more of an issue with abilityless-tools using this class, though.

Vaern:
I have only a few trivial nitpicks like localization but besides that it's an all-around stellar pr. I would merge this just for the gui, let alone the refactoring to make abilities nicer to work with :D

gradle.properties
abel1502: VS code already correctly reconginzes this file as UTF-8, so I don't think that's the problem. If you're referring to the replacement character in the name in line 14, that wasn't changed by any of my commits, and it is like that in the master branch already. I think the only change (other than my entry) is that CR-LFs seem to have been automatically converted into LFs at the ends of lines. I could revert it if that was intentionally that way, but I don't think it has any real effect either way

Vaern: weird, i guess it might be fixable on bob's end? though, not really an issue as you said

src/main/java/com/hbm/items/tool/ItemToolAbility.java
abel1502: Fair point, I'll fix this

abel1502: Should be fixed now

EDIT: Nope. Hold on...

abel1502: Alright, yes, this time it works

gradle.properties
abel1502: Huh. I looked at this file with the android github client, and here it actually show an æ turning into a replacement character in the diff. That's really weird, considering both the browser and vs code showed it as unchanged... I'll see what I can do about this

abel1502: Okay, I just retyped it by hand. I had previously assumed it was intentionally a replacement character there, so I'd dismissed it as a pun

HbmMods:
why does `ModItems` have 14,000 changes

abel1502:
@HbmMods it doesn't actually. The file is just too big for diff to figure out quickly that most things haven't changed. I get that effect every time I merge upstream into my local clone. I just open the merge view in VS code and let it think for half a minute, and then it admits there are 0 changes to begin with.

EDIT: Although it's the first time I've seen it on github, it was always a local thing before. I'll investigate

abel1502:
Nope, for me there's definitely nowhere near 14k changes

screenshot

abel1502:
Changed lines begin at 4470 and end at 4811, plus three lines of imports. The changes all boil down to replacing `.addBreakAbility` and `.addHitAbility` with `.addAbility` with corresponding arguments

abel1502:
@HbmMods pls merge soon, before more merge conflicts arise 🥺

PR #2158, opened 2025-05-20
the five hundredth crate dupe fix

MellowArpeggiation:
the reason this fix will work where others failed is because I did it ;)

PS: conveyor sorters no longer open GUI when placing conveyors, oops!


Menhyy:
GOAT

PR #2167, opened 2025-05-22
Parallelized ray explosion

mlbv:
Add optional parallel DDA explosion for major performance boost
- Explosion processing is around 20x faster on my setup (i7-14700KF, 20C/28T).
- Fully configurable, set `enableParallelization` to false to restore original behavior.
- Compatible with existing implementations.

Check the code for details.

Known Issue: The explosion shape differs slightly from the original algorithm.

### DIFF
--- a/src/main/java/com/hbm/config/BombConfig.java
+++ b/src/main/java/com/hbm/config/BombConfig.java
@@ -27,6 +27,7 @@ public class BombConfig {
 	public static int limitExplosionLifespan = 0;
 	public static boolean chunkloading = true;
 	public static boolean parallelization = true;
+	public static boolean accumulatedDestruction = true;
 
 	public static void loadFromConfig(Configuration config) {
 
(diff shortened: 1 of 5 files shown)


HbmMods:
Do you have a comparison image of the new crater shape?

mlbv:
It looked like a Reuleaux square and I'm working on a fix. I’ll push the changes soon and attach a comparison screenshot once it’s ready.

mlbv:
<details>
<summary>ExplosionNukeRayBatched</summary>

image
</details>
<details>
<summary>ExplosionNukeRayParallelized</summary>

image
</details>

Nuke used: *Little Boy* @ "Redstone Ready" superflat world

Its shape can be tuned in getEnergyLossFactor()

mlbv:
i got an even better idea

mlbv:
<details>
<summary>accumulatedDestruction</summary>

image
</details>

This one should be more physically accurate.

HbmMods:
Looks good, although I'm a bit unsure about the accumulated destruction part. Might not look as bad on regular terrain as it does on superflat, I'll make sure to run some tests on other world types too. In either case, I'll keep the config around, it's just a matter of whether it's on or off by default.

PR #2180, opened 2025-05-26
Geiger counter comparator integration

abel1502:
A comparator can now read the level of radiation a Geiger counter is detecting:
- 0 at exactly 0 rads/sec
- +1 per 5 rads/sec (the division is rounded up, so 2 rads/sec still gives a signal of 1)
- 15 at 75+ rads/sec

While making this, I noticed there was already ROR-reader compatibility, but I decided to pull through with it anyway. Considering the Geiger counter only really has one value, it makes sense to me to make it work with a comparator without the need for a radio frequency.

This was requested by `@gamerlancer` on discord.

### DIFF
--- a/src/main/java/com/hbm/blocks/machine/GeigerCounter.java
+++ b/src/main/java/com/hbm/blocks/machine/GeigerCounter.java
@@ -131,4 +131,22 @@ public class GeigerCounter extends BlockContainer {
 			return false;
 		}
 	}
+
+	@Override
+	public boolean hasComparatorInputOverride() {
+		return true;
+	}
+
+	@Override
+	public int getComparatorInputOverride(World world, int x, int y, int z, int side) {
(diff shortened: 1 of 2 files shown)


Kal19801:
The scope of this detection method might be a bit limited. Perhaps it would be better to allow the Geiger counter's block form to be right-clicked to open an interface for detailed configuration of redstone output behavior and thresholds.

abel1502:
@Kal19801 I think that would be overdoing it. Redstone-over-radio reader can already get the precise value in rads, which can then be handled by an ROR logic receiver in whatever way. This is deliberately meant as a simpler alternative. This is also in line with the other comparator-compatible blocks, such as inventories, and in those case this mod's approach is also to provide both a simple-approximate method (comparators) and a complex-precise one (storage level reader, or whatever it's called, and the ROR reader in general).

PR #2185, opened 2025-05-29
New OC functions for radar, launch pad and turrets.

Kirby7871:
Added getPos() to the radar, launch pad, and turrets.
This functions allows to fetch the coords of these machines.
(Per example now you are able to check where is the malfunctioning launch pad)


Kirby7871:
Also, yes, I reviewed this code in my own build of NTM and it works.

PR #2188, opened 2025-05-30
Automatic custom recipe sync for servers

MellowArpeggiation:
Does what tin says on: when enabled on servers (disabled by default), it'll automatically send any custom recipe configs to the client, allowing servers to change recipes without having to recreate modpacks, or even include the recipes in modpacks at all!

Does not affect any vanilla recipes like crafting tables or smelting, only our custom machines.

Also fixes:
* Detailed hitboxes for multiblock machines like chemplant, assembler, etc. now work for mouseclick interactions too, meaning you will no longer accidentally interact with your chemplant while trying to open a chest or similar
* Fix crash with crate NBT 6kb checking (the wrong guard was used


HbmMods:
image

PR #2194, opened 2025-06-02
QoL changes in bulk

abel1502:
I decided that instead of a hundred small PRs, I'll bundle the changes into a single big one. In summary:

- Wooden scaffolds can now be climbed (like a ladder) from all sides.
- Freestanding (freehanging?) chains, vines and any other `isLadder()` blocks without a collider can now be climbed by holding the spacebar. Ladders with colliders are unaffected.
- Strand caster timeout now counts from either the last cast or the last pour. This results in more organic handling of continuous casts. Previously, a caster which had been idle for a while would become primed and cast the first poured material prematurely. Now this should no longer be the case. Long pours also behave more naturally, with the caster mostly pouring full loads, except at the very end.
- Meteorite blocks no longer randomly drop angry metal. Since its use has effectively been removed (the defabricator is no more, and the eye of harmony is deprecated along with the AMS), I thought it didn't make sense to have it anymore. If needed, it can still be crafted in an assembler.
- Pattern slots have been slightly refactored.
- The automatic buzz saw now ignores immature plants and automatically replants any that drop seeds. (The headcanon for this is that seeds get replanted naturally when they fall onto the ground).
  - This required a rework of the scanning algorithm, from raytracing to grid-scanning. Performance-wise, the new approach is about the same as the previous one (now, a single 300-iteration loop is made, most of the iterations are skipped automatically; before, the raytracing took 200 steps along the vector). In practice, there is no noticeable performance difference. 
- The template folder now supports scrollwheel to cycle through the pages.
- Mass storage input and output slots are no longer bottlenecked by entity tick rate or network latency. You can now shift-click an entire inventory into a mass storage, or take stacks out with an autoclicker.
  - Sometimes this results in a graphical glitch: if shift-clicking too many stacks into a mass storage with output enabled, the output slot appears vacant. It can still be taken from, and its contents correctly appear when the UI is closed and opened again. I couldn't establish what causes this, but it doesn't seem serious and only appears very occasionally, so I don't think it's a big issue.
- Storage crates (and the safe) can be named in their item form, and retain the name when placed. It is shown in the UI and in an overlay when looking at the block. Unnamed crates are unaffected.
- Crates that have been crafted and those that have been placed and harvested can be stacked together (provided there is no other difference). Previously, they differed in the `spiders` NBT field -- one had none, the other had `false`.
- Renaming items in an anvil no longer increases their repair cost. This has been implemented in vanilla 1.8, so it may be considered a backport. The rationale is to maintain NBT compatibility between a crate (or any other block) that has been renamed and itself after being placed and harvested. Thanks to this change, they can be stacked together properly.

See extended commit messages for more in-depth justifications in some cases.

As an honorable mention, while not part of the PR, I also [fixed](https://github.com/GTNewHorizons/NotEnoughItems/pull/639) an [issue](https://github.com/GTNewHorizons/NotEnoughItems/issues/638) in GTNH's NEI, which was breaking drag-and-drop into NTM's filter slots.

By the way, are there any criteria for getting a bobblehead, or may I just make one for myself at this point?

Update: a couple more changes:

- Now two opposing fans push entities into the middle, as opposed to not affecting them at all. This is done by making the fan force decrease over distance, down to half of its original value at 10 blocks. This also means a fan can now be overpowered at a distance. If that is undesirable, the base force could be increased.
- The autosaw now respects walls. If it hits a solid wall (wood that gets destroyed doesn't count), it immediately retracts back. To account for the case where a plant might be inaccessible behind a wall, after a saw hits a wall, it ignores any potential targets for the next 5 ticks (5 degrees of rotation).
- The autosaw can now be suspended for maintenance using a screwdriver.
- The fan mode can be toggled between the old and new one with a hand drill.


Vaern:
> By the way, are there any criteria for getting a bobblehead, or may I just make one for myself at this point?

the criteria is Nepotism but your work has surpassed the threshold about three times over :D

Ashen-sysadmin:
Renaming items not increasing cost is a good idea, but also a fix for vanilla behaviour which is potentially outside of the scope of ntm? Same with altering vine behaviour.
Ask other people first though, I'm not an authority on any of this. It's otherwise an excellent batch of fixes and features.

DrIshirozaki:
This is great

abel1502:
@MerrittK fair, but I couldn't find an existing standalone mod for this version of Minecraft for either of those fixes. Still, if there's any problem with them, they could be locked behind configs.

Also, according to the readme:

> NTM has certain behaviors intended to fix vanilla code or to increase compatibility in certain cases where it otherwise would not be possible. These behaviors have the potential of not playing well with other mods, and while no such cases are currently known, here's a list of them.

So in general, it seems fixes are considered acceptable for the mod. But Bob obviously has the final word

abel1502:
Updated with another small feature, see first post

abel1502:
And another addition. I'll probably stop writing new comments every time, you can see the commits and the "Update" section in the core post

gmhs022:
The Eye of Harmony is still used as a core for the DFC, all the AMS cores were transferred over to it.

Banger changes overall though, especially the fan one, will make smaller buzzsaw-based farms way easier to set up 🙏

abel1502:
@gmhs022 Oh, I didn't know that. But as I also said, the angry metal can still be made from the meteorite block in an assembly plant. I guess it would only make sense to keep the angry metal drop if it was intended as a punishment for not using silk touch, but then it might as well not drop anything

HbmMods:
1. The buzzsaw was never supposed to be used on crops to begin with, you would have to put it in the floor, where it covers a circular area (meanwhile watered soil covers a square) and you're basically forced to use an external fuel source. I do plan on adding a different machine for that.
2. Crates having their names show as a tooltip is nice I guess but feels a bit redundant considering you can already do that with signs. Either way, I feel like the current implementation might break quite quickly on servers. If memory serves right, inventory names are sent with the packet that opens the GUI, which means that until the GUI is opened at least once, the client probably doesn't really have a way of knowing what the name is beforehand.
3. Mass storages aren't really meant to be used manually all that much, and I'd rather we don't introduce syncing issues like that.
4. Concerning the anvil cost thing, I'm not sure if we should alter vanilla behavior like that, considering it doesn't have anything to do with NTM, unlike most of the other compatibility changes to vanilla stuff we have. I mean I'm not really against it, but it feels like a weird thing to include.
5. Diminishing fans are nice for things where opposing fans are needed, but less nice when not, for example elevators and wind tunnels. Diminishing fans could be the default, but I'd rather the original way were toggleable via screwdriver.
6. Toggling the buzzsaw would probably make more sense via redstone signal, you could automate short durations of operation so the thing doesn't run on idle for as long while waiting for trees to grow.

abel1502:
@HbmMods alright, I'll work on this today or tomorrow. Some comments:

1. A different machine would be nice, but a buzzsaw works quite well for this scenario for now. Besides, with the changes proposed, a buzzsaw placed in the floor would still work with saplings (since a sapling can accept bonemeal, it is considered immature), so self-sustaining operation with crops is very much possible.

2. Will investigate and fix if necessary. Tooltips are superior to signs because they don't obstruct the clickable area. In my previous survival stockpile, I had to give up on signs on crates because of that, and instead rely on color-coded patterns on the stockpile floor. EDIT: Actually, I just realized the custom name isn't even saved to NBT, and not only for crates, but for everything. Does it just vanish on a server reload? This might need a larger fix than I thought

3. From my survival playthrough experience, manual use of mass storage is almost inevitable (at least until I implement the warehouse idea I suggested on discord). But maybe the same could be implemented with take-all and store-all buttons, instead of altering the slot behavior. Would you prefer that?

4. I agree that it's weird, but honestly, so is the vanilla behavior. In vanilla, you'd only really rename gear, but with mods adding portable inventories (or vanilla adding shulker boxes), you have a good reason to rename blocks. But repair cost for a block makes very little sense, and, as I mentioned, interferes with stacking in a very inconvenient way. I could, if you'd like, restrict the effect only to this mod's blocks, as to not overstep the boundaries of the mod's effect.

5. Will do.

6. It was the way I was considering, but then I realized that in a center-funnel design farm, there's nowhere to input a redstone signal from. Especially considering an RoR torch or a lever cannot be placed on the saw directly. Also, I feel like the more dominant use case for disabling a buzzsaw is fixing the farm or replanting something, in which case a manual action seems more convenient. For automation, I think a better solution would be reducing the saw's internal tank to something like 1 or 2 mb, so that cutting off the pipe would shut down the saw almost immediately. That way, an automatic switch can be implemented with a redstone valve.

abel1502:
@HbmMods fixed 2, made sure tooltips work on servers. Also made so that `customName` is saved to NBT in every block that has it -- previously it would reset upon world reload. Will do fan force falloff toggle next. Waiting for feedback about reducing buzzsaw oil buffer and restricting the anvil fix to this mod's blocks.

abel1502:
Made fan mode toggleable with a hand-drill (point 5). Not a screwdriver because that was already used to cycle its orientation. I'm out of time for today, and I'm awaiting feedback for the rest of the suggested improvements

PR #2210, opened 2025-06-11
Slight change to ability GUI handling

abel1502:
The new single-button approach felt weird, so I made it function as a modifier to the "cycle ability" button instead. I think it feels more natural this way, as well as being consistent with shift-clicking to reset.


HbmMods:
nah

PR #2212, opened 2025-06-12
Metal trapdoor variations

abel1502:
It always bugged me that I had to cap off cool metal ladders with a lame vanilla wooden hatch with holes in it. Well, no more. This PR adds corresponding trapdoor variants for each of this mod's stairs. Crafted with 1 regular trapdoor and 1 ingot of material. As a bonus, the new trapdoors are climbable if they are at the top of a ladder, like in newer vanilla versions.

Also, I'm not great with textures, so the colors might be a bit off. Tell me if anything looks particularly egregious, I'll try and fix it.


abel1502:
Update: made the bunker structure use a sturdy trapdoor for its hatch, since it's already using sturdy ladders

HbmMods:
Honestly nah. Most of them don't look good, there's four with almost identical color, and it's a ton of block IDs used for something that's just not that great. Ladders have the same issue btw, I regret adding way too many of them.

abel1502:
@HbmMods If I only leave the steel one, would that be fine? I too think having so many variants is excessive, but I feel that at least some kind of blast-resistant-ish and climbable hatch would be almost indispensable

PR #2234, opened 2025-06-25
Tantalum 2

70000hp:
### **Reworked tantalum chain:**
First step replaces peroxide with sulfuric acid, and removed fluoride requirement due to the sole way of effective production at scale being bedrock ore
image

Second Step has been completly reworked:

- Now uses Ethanol, cleaned coltan, and one of two routes
- - Aluminium powder 
image
- - Liquid Sodium
image

Aluminium can be mass produced effectively via bauxite, however mass sodium production is difficult until later game, where a sodium sink is needed

Last step now also has a Vitriol byproduct, and an alternative electrolyzer option
image

image

**Other Changes**
Ethanol has obtained some slight improvements and a pre-liquefier, inefficient recipe from hemp in the Comb oven, and all sugar cane related ways have been buffed
Optional, improved circuit/chips recipes have been added using tantalum, primarily for late game circuits  (recipes not available on 528)
An alternate Coltan bedrock ore now can spawn in the nether, however requires chlorine to be mined
- I wonder where one can get chlorine from in the nether


Ashen-sysadmin:
Nifty. - This brings Tantalum to modern standards.

HbmMods:
whatever happened to "run major gameplay changes by bob"

PR #2235, opened 2025-06-25
Shredder Preferential Ore registry.

Ashen-sysadmin:
Inspired by various other mods that do it, I tried my hand implementing a preferential config for autogen recipes. 

Currently applies to the shredder, now it will produce NTM dusts by default, but the config could be changed to say: gregtech dusts, then ntm dusts, with enderio as a fallback if a modpack maker wanted.

I also accidentally fixed advanced rocketry moon turf compat - because of copy paste typo's, it had been set to only register the recipes if galacticraft was installed.


Ashen-sysadmin:
"Admitting to a crime is something I 100% will do on any discord server" - someone apparently

PR #2239, opened 2025-06-27
fixing dickhouse glitched

Lazzzycatwastaken:
self explaantory
and also buffs loot (adds a flaregun + flares)


Vaern:
fuck _I_ was going to make a structure with a flaregun stack

HbmMods:
what does that even fix

BallOfEnergy1:
IIRC it fixes the issue with paintable power cables not having the right block in the structure.

Lazzzycatwastaken:
what gamma said @HbmMods

HbmMods:
no it doesn't

Lazzzycatwastaken:
Fym “it doesnt” i legit rebuilt the thing WITHOUT the block entirely

HbmMods:
i mean it doesn't exactly fix the underlying issue but whatever

PR #2241, opened 2025-06-28
QoL tweaks 2

abel1502:
I realized the year-month naming scheme was a dumb idea because I'll definitely shit out more than one of these per month. This time includes the following changes:

- Spotlights are no longer considered full blocks by entity pathfinding. This means no more zombies jumping and spinning around on spotlights on the floor like idiots. While they were already intangible for entities, pathfinding relied on material properties by default, which was iron in this case.
- Spotlights are transparent on the map, similar to torches. Same root cause as the previous point.
- Large doors can now be clicked through when open.
- The crate GUI only locks the held item slot if the crate is actually opened from an inventory. The previous solution locks whatever item is in the player's hand, even when they open a crate that is placed in the world.
- There is now a flat version of the AoE tool ability, for when you don't quite need to excavate a full NxNxN cube, but an NxNx1 plane would suit your needs just right.
- Tool abilities that break multiple blocks now position all the drops in the place of the original target block. The previous implementation (every block drops where it's broken) was mostly fine with a magnet, except for when veinminer broke a block that was completely encased on all cardinal directions through a diagonal. In that case, the magnet would actually cause an annoying visual glitch, where the dropped block would fly all over the screen client-side, but not get picked up server-side until you found and approached its prison. The new behavior solves it, and is generally consistent with veinminer mods out there.
- Tool abilities that break multiple blocks only apply exhaustion for the first targeted block, instead of (implicitly) for every block mined. Paying with extra hunger would make sense if the ability was inherent to the player, but since it's supposed to be a benefit of a more advanced tool, this made more sense. Durability/charge on the tool is still consumed for every block.

There's another little thing, but I thought it might be controversial, so I didn't include it here. Tell me if I should.
Asphalt-covered concrete, effectively an upper slab of asphalt and a lower slab of concrete in the same tile. Gives the speed boost of the asphalt, but doesn't result in awkward black stripes on ceilings when used on the second floor and above. Crafted from a block of asphalt above a block of concrete, giving two in return.

All was tested and works, obviously. I might add something else while this is pending. If I do, it will be listed below. Not a draft PR though, what is here is ready to be merged.


HbmMods:
1. good
2. good
3. good
4. good
5. I don't really see how that's useful to the point where it deserves its own area mode. If you want extra precision like that you really shouldn't be using a large area tool to begin with.
6. Not really a fan of doing it like that but sure
7. The entire idea was to speed up mining with a proportional tradeoff, that being the tool breaking and hunger increasing at adjusted rates. I feel like expecting players to eat just a bit more for way faster mining (18x in the case of AoE 1) is more than justified.

On another note, can we not do grab bag PRs like that, it just makes reviewing it and getting it into a mergable state way harder, and the literal first thing the contribution guidelines say is to not do that.

abel1502:
Alright, noted, sorry. I'll separate stuff from now on.

5 is useful for, for example, replacing large sections of walls / floors. Veinminer is messy and dangerous, and conventional AoE would break half the stuff inside the room along with the wall. And going block-by-block would be 25 times slower than AoE 2, which is significant.

7 I'll revert now, gimme a sec

PR #2245, opened 2025-06-28
Make conveyor extractors take items to inventory when not facing a conveyor

abel1502:
This allows using them as an ad-hoc hopper with a small buffer. Previously their buffer was effectively rudimentary -- it was used for output, but nothing ever inputted into it.

### DIFF
--- a/src/main/java/com/hbm/tileentity/network/TileEntityCraneExtractor.java
+++ b/src/main/java/com/hbm/tileentity/network/TileEntityCraneExtractor.java
@@ -10,6 +10,8 @@ import com.hbm.items.ModItems;
 import com.hbm.module.ModulePatternMatcher;
 import com.hbm.tileentity.IControlReceiverFilter;
 import com.hbm.tileentity.IGUIProvider;
+import com.hbm.util.InventoryUtil;
+
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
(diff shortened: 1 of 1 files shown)


HbmMods:
The buffer's purpose is being able to receive items from other sources (Thermal pipes, pneumatic tubes, assembler output chutes) and turn them into conveyor belt items. Making extractors pull items into the buffer just feels like a more expensive version of hoppers, at which point it's just easier and more effective to use the conveyors the way they were intended to be used.

abel1502:
@HbmMods not exactly. Extracting with just an extractor means the machine shuts down if it overflows, while with a conveyor, it keeps going and the items either spill or get deleted depending on the corresponding inserter's config.

And honestly, a more expensive hopper that doesn't have to be placed on the bottom, with more slots, a filter and support for speed upgrades sounds like a useful thing on its own, if you think about it.

HbmMods:
Conveyors can be shut off with a comparator, using RoR that's easy to do and works over any range.

abel1502:
@HbmMods true, but now you're talking a 5-block machine for a quick-and-dirty output buffer. I agree that this change has no use in a permanent organized production line, but for simply expanding the input and output buffers of an individual machine, this is perfect. Also, the new extractors can be chained to create a larger output buffer, which actually can be of use even in a permanent factory with RoR-driven signaling.

PR #2246, opened 2025-06-29
Turn burner & electric press into proper pseudomultiblocks

abel1502:
They seem to have been implemented largely based on copy-pasted furnace code, and seemingly predate `BlockDummyable`. That meant the top two blocks of a press were intangible. Now they're in line with the rest of the machines, including the conveyor press.

I also added a bit of compatibility handling to `BlockDummyable`. It can now identify legacy monoblocks when deleting a suspected orphan, and promote them to lone cores. This makes this change savegame-compatible. Existing presses will remain single-block-sized and functionally identical to before. If broken and replaced, they will be 3-block-tall. The solution is general (the updated block must override `isLegacyMonoblock` to detect if it is a correct legacy block. There is a caveat, though: the fix is only triggered when the block is updated, which doesn't happen on world load, unfortunately. The best workaround for this that I came up with was triggering a block update from the tile entity tick if the metadata for the block indicates it's a dummy. The performance cost is negligible, thankfully.

### DIFF
--- a/src/main/java/com/hbm/blocks/BlockDummyable.java
+++ b/src/main/java/com/hbm/blocks/BlockDummyable.java
@@ -103,7 +103,12 @@ public abstract class BlockDummyable extends BlockContainer implements ICustomBl
 		ForgeDirection dir = ForgeDirection.getOrientation(metadata).getOpposite();
 		Block b = world.getBlock(x + dir.offsetX, y + dir.offsetY, z + dir.offsetZ);
 
-		if(b != this) {
+		// An extra precaution against multiblocks on chunk borders being erroneously deleted.
+		// Technically, this might be used to persist ghost dummy blocks by manipulating
+		// loaded chunks and block destruction, but this gives no benefit to the player,
+		// cannot be done accidentally, and is definitely preferable to multiblocks
+		// just vanishing when their chunks are unloaded in an unlucky way.
+		if(b != this && world.checkChunksExist(x - 1, y - 1, z - 1, x + 1, y + 1, z + 1)) {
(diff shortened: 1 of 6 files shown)


70000hp:
I think that by now them not being proper multiblocks is deliberate?

abel1502:
@70000hp considering their code used mojang-obfuscated variable names, I'm pretty sure it just wasn't touched for a long while

abel1502:
I think there might be an issue with this, marking as a draft while I'm investigating. I just saw a lot of multiblocks have disappeared from my world. In all likelyness, it's probably a consequence of one of the intermediate attempts, considering it doesn't seem to happen anymore, but better safe than sorry. I'll undraft when I'm sure it's safe

abel1502:
Okay, I think I found the problem. I'll test my fix ~~tomorrow~~ right now.

Basically, a dummyable on a chunk border might wrongly consider itself an orphan if its neighbor is in an unloaded chunk. It's weird that the problem didn't show up before, as it was present before my changes. I suspect one of my intermediate solutions must've aggravated its effect.

abel1502:
Okay, yes, I can confirm the problem no longer occurs. Removing the draft status

MellowArpeggiation:
vetoing this one, removes ability to squish PU bobblehead and plushies

70000hp:
> vetoing this one, removes ability to squish PU bobblehead and plushies

seconded

abel1502:
> vetoing this one, removes ability to squish PU bobblehead and plushies

Compelling argument. Might make it unmultiblockable with a handdrill or something

abel1502:
@MellowArpeggiation @70000hp here ya go

PR #2247, opened 2025-06-30
Add a steel trapdoor

abel1502:
Basically #2212, but with only the steel trapdoor. I feel that at least some kind of blast-resistant & climbable trapdoor would fill a long-running gap in the mod.


ArfGg57:
Yeah steel trapdoors would be so nice. I wouldn't have to use the crappy vanilla trapdoor anymore lol.

PR #2255, opened 2025-07-02
Tier 4 Balefire missile

FatSassin:
Adds a  new tier 4 missile, balefire missile. Comes with an amateur retexture of the nuclear missile.


HbmMods:
nah

PR #2261, opened 2025-07-04
Dungeon Action Block, and Mob Decorator improvments

70000hp:
**Changes**

- Reworked the Dungeon Spawner to be a multi-purpose logic block for structures:

   - Added a structure wand block (similar to the ones used for loot tables or jigsaws), which can determine:
     - Actions, which are ran every tick
     - Conditions, which, when true, will increment an internal phase counter, and reset the internal timer
     - Interactions, which happen when a player right clicks the block, and has all the information a right click gives
     - Disguises, a way to make the dungeon action block look like another block, following the same rules as paintable cables
     
   - This also involves a significant rewrite of previous code, but also comes with fixes for any previous structures that used it
- Changed mob decoration code
   - The old system was inefficient (several hashmaps and arrays created per decorated mob) and impossible to use by any process outside of the vanilla mob decoration event, and lacked ways to extend it
   - Relevant code was moved to it's own utility class, MobUtils, which allows other things, such as the aforementioned dungeon spawners, to decorate mobs, and for multiple mob decoration pools to be made for structure purposes
   - Slightly changed both gun mob, and mob armor loot tables, due to some small balancing issues (full security armor zombies, gun mob scaling working poorly and not scaling with how difficult x soot is to produce), can be reverted if needed


70000hp:
https://github.com/user-attachments/assets/955c2196-c675-47c2-9d45-36d417b1edb3

https://github.com/user-attachments/assets/41b35409-75a4-438a-b1d3-a2761a9540ee


https://github.com/user-attachments/assets/cc7f3e47-626e-4465-81ac-ec92c950b74d


https://github.com/user-attachments/assets/a11510ed-be15-4c43-8f6a-f75e07c26756


Examples of what it can do

TreeVirus:
slop

FatSassin:
Epic

Lazzzycatwastaken:
plase 🟢 i NEED thixss

HbmMods:
next time make your own fucking block

BallOfEnergy1:
the mob block is here

HbmMods:
Ok no hold on, you're reading component IDs from NBT, resolving them and then running those without any null checks, so any issue with NBT or shift in IDs is gonna cause this thing to crash.

src/main/java/com/hbm/world/gen/util/DungeonSpawnerActions.java
```diff
@@ -0,0 +1,261 @@
...
+				TileEntity te = null;
+				if(block instanceof  BlockDummyable){
+					int[] coreCoords = ((BlockDummyable) block).findCore(world,checkX,checkY,checkZ);
+					te = world.getTileEntity(coreCoords[0], coreCoords[1], coreCoords[2]);
+				}
+
+				if (te instanceof TileEntityDoorGeneric) {
+					TileEntityDoorGeneric door = (TileEntityDoorGeneric) te;
+					door.setPins(456);
+					door.close();
```
HbmMods, line 235: yeah so basically that's not a thing that exists in `TileEntityDoorGeneric`

PR #2269, opened 2025-07-09
The Logic Block

70000hp:
exactly the same PR as before, but the dungeon spawner is left untouched, and half of it is renamed


Ashen-sysadmin:
Why are there dupe commits?

legendarydoge30:
evillarry

BlitzFan2059:
bob plz accept

Lazzzycatwastaken:
did you remove the fucking method that doesnt exist

Issue #2266, opened 2025-07-08
Suggestion: Add Forge Microblocks compatibility


fbfnysnshnsgnwg:
### Describe the bug

If I had forge microblock compatibility, I would be able to cover up redstone set up with project red, which is useful as it can go on walls and ceilings.

For example there's redstone back here and I need to cover it up with a slab, otherwise it looks like this: 

<img width="1920" height="1052" alt="Image" src="https://github.com/user-attachments/assets/3453521a-7cea-43da-b7a6-aa40b7100b6a" />

PR #2270, opened 2025-07-09
Added Forge MicroBlocks support

Toshayo:
I randomly opened issues for no reason and saw one feature (https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/issues/2266) I've already implemented in a separate mod for a modpack

### DIFF
--- a/src/main/java/com/hbm/handler/microblocks/MicroBlocksCompatHandler.java
+++ b/src/main/java/com/hbm/handler/microblocks/MicroBlocksCompatHandler.java
@@ -1,0 +1,77 @@
+package com.hbm.handler.microblocks;
+
+import com.hbm.blocks.ModBlocks;
+import cpw.mods.fml.common.event.FMLInterModComms;
+import net.minecraft.block.Block;
+import net.minecraft.item.ItemStack;
+
+public class MicroBlocksCompatHandler {
+	public static void preInit() {
+		registerMicroBlockCompat(ModBlocks.basalt);
+		registerMicroBlockCompat(ModBlocks.basalt_smooth);
+		registerMicroBlockCompat(ModBlocks.basalt_brick);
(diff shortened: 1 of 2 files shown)


HbmMods:
Non-intrusive IMC-based compatibility to the price of [none]?? Satisfactory.

abel1502:
I had FMB compat in mind for a while, though I was thinking making grates & steel walls treated as microblocks so that multiple could coexist at once. Though that would probably be a bit more intrusive, regrettably

abel1502:
... Or it might be as trivial as adding [this](https://github.com/GTNewHorizons/Applied-Energistics-2-Unofficial/blob/ab5f3722223a50edfdcc22725f5c7df336e9c7c8/src/main/java/appeng/block/networking/BlockCableBus.java#L77-L85) and letting FMB do the rest -- apparently it works for GTNH AE2.

Toshayo:
> I had FMB compat in mind for a while, though I was thinking making grates & steel walls treated as microblocks so that multiple could coexist at once. Though that would probably be a bit more intrusive, regrettably

I'm not a user of microblocks, so i cannot say for sure, but microblocks doesn't add non full cube blocks support, does it? Or only if it is handled correctly?

abel1502:
> I'm not a user of microblocks, so i cannot say for sure, but microblocks doesn't add non full cube blocks support, does it? Or only if it is handled correctly?

Not through the usual way, unfortunately, no. But it is possible for a mod to register one of its blocks as a multiblock part (without any additional cutting). This is how AE2 cables/buses can be placed in the same block with FMB covers, for example. I'm not exactly sure what is the minimum viable implementation for this, though. AE2 seems to imply the Microblocks mod does everything automatically if it detects a specific field in the block type using reflection, but then they also have a bunch more compat code elsewhere.

PR #2281, opened 2025-07-13
Smarter shift click for schrabidium transmutator & strand caster

abel1502:
Correctly handles shift-clicking batteries, redcoil/euphemium capacitors and molds into the schrabidium transmutator and the strand caster


HbmMods:
both of these are gonna die like really soon anyway

abel1502:
@HbmMods The strand caster too?

PR #2337, opened 2025-08-05
Continuation of russian translation

Bufka2011:
yep


Bufka2011:
I did this because the previous translator (MrKimkimora, if I'm not mistaken) evaporated.

PR #2361, opened 2025-08-15
QMAW russian translation and ru_RU.lang update

Bufka2011:
Yep


HbmMods:
i assume that means that the localizations actually work because i didn't actually test anything besides `en_US` yet

PR #2365, opened 2025-08-17
More custom machine options

FOlkvangrField:
Now we can have custom machines render custom models and play custom sounds when running recipes, and custom machines now support hot reloading.


src/main/java/com/hbm/config/CustomMachineConfigJSON.java
```diff
@@ -231,8 +254,33 @@ public static void readConfig(File config) {
...
+						model_Bounding_x2 = model_Bounding_x2 < compDef.x ? model_Bounding_x2 : compDef.x;
+						model_Bounding_y2 = model_Bounding_y2 < compDef.y ? model_Bounding_y2 : compDef.y;
+						model_Bounding_z2 = model_Bounding_z2 < compDef.z ? model_Bounding_z2 : compDef.z;
+					}
+				}
+
+				if(machineObject.has("customModel")) {
+					JsonObject modelObject = machineObject.get("customModel").getAsJsonObject();
+					MachineConfiguration.CustomModel customModel = new MachineConfiguration.CustomModel();
+					customModel.customModel = new HFRWavefrontObject(new ResourceLocation(RefStrings.MODID, modelObject.get("model").getAsString()));
```
HbmMods: Instantiating an obj model requires minecraft's resource loader which is unavailable on servers (`Minecraft.getMinecraft().getResourceManager().getResource(resource)`), so I expect this to crash

src/main/java/com/hbm/tileentity/machine/TileEntityCustomMachine.java
```diff
@@ -192,6 +196,22 @@ public void updateEntity() {
 							this.processRecipe(cachedRecipe);
 							this.cachedRecipe = null;
 						}
+						if(this.progress > 0) {
+							if (audio == null) {
+								audio = this.createAudioLoop();
```
HbmMods: This is part of `!worldObj.isRemote` which runs serverside, calling `getLoopedSound` on the `ServerProxy` returns `null` which *will* crash

src/main/java/com/hbm/tileentity/machine/TileEntityCustomMachine.java
```diff
@@ -214,6 +234,22 @@ public void updateEntity() {
 								this.useUpInput(recipe);
 								this.processRecipe(recipe);
 							}
+							if(this.progress > 0) {
```
HbmMods: Why does this exist twice? Wouldn't it be easier to set a flag in both cases, and then play the sound based on that?

src/main/java/com/hbm/config/CustomMachineConfigJSON.java
HbmMods: Additionally, models have the disadvantage that there is no default fallback model when the path is incorrect or the model format isn't compatible, so the game will crash if that's the case. Ideally, a config should be able to handle things like that without breaking entirely.

HbmMods (changes requested):

FOlkvangrField:
Should have fixed the problems above @HbmMods

src/main/java/com/hbm/render/tileentity/RenderCustomMachine.java
```diff
@@ -1,50 +1,97 @@
...
+
+				}
+
+				SmallBlockPronter.draw();
+
+				GL11.glPopMatrix();
+			}
+			else if(config.customModel!=null){
+				try{
+					IModelCustom customModel = new HFRWavefrontObject(new ResourceLocation(RefStrings.MODID, config.customModel.customModel));
```
HbmMods: You're instantiating a new model instance on every frame, I don't think that's a good idea. Instead it would be better if the config held a cached model that is only created once, the first time it is rendered.

src/main/java/com/hbm/tileentity/machine/TileEntityCustomMachine.java
```diff
@@ -160,12 +165,12 @@ public void updateEntity() {
...
 				for (FluidTank tank : this.outputTanks)
 					if (tank.getFill() > 0)
-						this.sendFluid(tank, worldObj, pos.getX(), pos.getY(), pos.getZ(), pos.getDir());
+						this.tryProvide(tank, worldObj, pos.getX(), pos.getY(), pos.getZ(), pos.getDir());
 				this.sendSmoke(pos.getX(), pos.getY(), pos.getZ(), pos.getDir());
 			}
 
 			if (this.structureOK) {
-
+				float volume = 1F;
```
HbmMods: This is still serverside so it does nothing. Volume is never anything other than 1 here.

src/main/java/com/hbm/tileentity/machine/TileEntityCustomMachine.java
```diff
@@ -223,6 +228,24 @@ public void updateEntity() {
 				this.progress = 0;
 			}
 			this.networkPackNT(50);
+		} else {
+			float volume = this.getVolume(1F);
+			if(this.progress > 0) {
```
HbmMods: Also perhaps not the best way of doing this, every recipe cycle there is one tick where the machine is running, but the progress is 0, so this would momentarily stop the sound and reset it every time the machine finishes a recipe, even if it is still running.

src/main/java/com/hbm/tileentity/machine/TileEntityCustomMachine.java
```diff
@@ -262,6 +285,31 @@ public void deserialize(ByteBuf buf) {
 		this.matcher.deserialize(buf);
 	}
 
+	@Override
+	public AudioWrapper createAudioLoop() {
+		return MainRegistry.proxy.getLoopedSound(config.progressSound, xCoord, yCoord, zCoord, 1.0F, 10F, 1.0F);
```
HbmMods, line 292: What will happen if the config has no sound set at all? It will print a log message when it can't play the sound, and assuming that an invalid sound never fulfills `audio.isPlaying()`, then it'll just try to restart every tick, failing, and logging the missing sound.

HbmMods (changes requested):

src/main/java/com/hbm/tileentity/machine/TileEntityCustomMachine.java
FOlkvangrField: Does "config has no sound set at all" means that the player forgot to write the 'progressSound' option in the config file, or that the player wrote a wrong resource path in the 'progressSound 'option?

FOlkvangrField: > What will happen if the config has no sound set at all? It will print a log message when it can't play the sound, and assuming that an invalid sound never fulfills `audio.isPlaying()`, then it'll just try to restart every tick, failing, and logging the missing sound.

Does "config has no sound set at all" means that the player forgot to write the 'progressSound' option in the config file, or that the player wrote a wrong resource path in the 'progressSound 'option?

HbmMods: Either possibility would likely cause that scenario. According to the config, the sounds are optional, so it needs to be able to handle having no sound at all.

FOlkvangrField: > Either possibility would likely cause that scenario. According to the config, the sounds are optional, so it needs to be able to handle having no sound at all.

So should other optional options such as "maxHeat" also be modified to prevent the above situation?

FOlkvangrField: Should have fixed these problems @HbmMods

src/main/java/com/hbm/config/CustomMachineConfigJSON.java
```diff
@@ -46,7 +57,10 @@ public static void initialize() {
 			writeDefault(config);
 		}
 
+		customModels.clear();
 		readConfig(config);
+		registerCustomModels();
```
HbmMods: `initialize()` is called on all sides, but `registerCustomModels()` and `doesSoundExist()` are clientside only which will crash servers instantly as soon as the custom machine config loads. To run client things without breaking the server, make a new empty method in `ServerProxy`, then override that same method in `ClientProxy`, add your clientside only calls there, then call that method via `MainRegistry.proxy`.

HbmMods (changes requested):

src/main/java/com/hbm/config/CustomMachineConfigJSON.java
FOlkvangrField: > `initialize()` is called on all sides, but `registerCustomModels()` and `doesSoundExist()` are clientside only which will crash servers instantly as soon as the custom machine config loads. To run client things without breaking the server, make a new empty method in `ServerProxy`, then override that same method in `ClientProxy`, add your clientside only calls there, then call that method via `MainRegistry.proxy`.

Like the fix 3 below？

FOlkvangrField:
Bob, it has been a month, do you forgot this? @HbmMods

FOlkvangrField:
@HbmMods bob？

Bufka2011:
@FOlkvangrField there is a change requested

HbmMods:
thanks for reminding me

PR #2368, opened 2025-08-19
i18n for RBMK DODD

RealSilverMoon:
image

More user-friendly for **_human_** recognition now!


Bufka2011:
You should merge this pull request! Now!

HbmMods:
Not a fan of heat, yield and xenon being lowercase, nor there being *two* values called "heat" with no visible distinction between them. One refers to the column's heat stat and the other is the actual fuel item's heat values.

RealSilverMoon:
> Not a fan of heat, yield and xenon being lowercase, nor there being _two_ values called "heat" with no visible distinction between them. One refers to the column's heat stat and the other is the actual fuel item's heat values.

Yeah that makes sense

PR #2371, opened 2025-08-20
Fixed the multilingual feature failure error of the "QMAW" system

FOlkvangrField:
Fixed the multilingual feature failure error of the "QMAW" system;
Force reading JSON files with custom text in UTF-8 encoding to avoid 'FileReader' reading JSON files with system default encoding other than UTF-8, which may cause garbled custom text.


arantirnecrolord:
bob, wake up🔔🔔🔔

PR #2384, opened 2025-08-23
Update Simplified Chinese Localization

Creeper-banner:
Updated zh_CN.lang & QMAW Chinese localization (tested with PR #2371 and worked fine).
Some of the QMAW localization is based on @NarekoMichigami810 's works.


HbmMods:
The changes done in #2374 seem to conflict with these due to editing many of the same files, I can't merge this PR and github's conflict resolve doesn't seem to work.

PR #2389, opened 2025-08-24
ME Storage bus compat for Arc Furnace

R-Kaenbyou:
also have some improvements for arc furnace shift-clicking

### DIFF
--- a/src/main/java/com/hbm/handler/ae2/AE2CompatHandler.java
+++ b/src/main/java/com/hbm/handler/ae2/AE2CompatHandler.java
@@ -14,5 +14,6 @@ public class AE2CompatHandler {
     @Optional.Method(modid = "appliedenergistics2")
     private static void registerHandler() {
         AEApi.instance().registries().externalStorage().addExternalStorageInterface(new MSUExternalStorageHandler());
+		AEApi.instance().registries().externalStorage().addExternalStorageInterface(new AFLExternalStorageHandler());
     }
 }
--- a/src/main/java/com/hbm/handler/ae2/AFLExternalStorageHandler.java
+++ b/src/main/java/com/hbm/handler/ae2/AFLExternalStorageHandler.java
@@ -1,0 +1,34 @@
+package com.hbm.handler.ae2;
+
+import com.hbm.tileentity.machine.TileEntityMachineArcFurnaceLarge;
+import com.hbm.tileentity.TileEntityProxyCombo;
+import cpw.mods.fml.common.Optional;
+
+import appeng.api.networking.security.BaseActionSource;
+import appeng.api.storage.IExternalStorageHandler;
+import appeng.api.storage.IMEInventory;
+import appeng.api.storage.StorageChannel;
+import appeng.me.storage.MEMonitorIInventory;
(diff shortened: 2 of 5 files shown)


HbmMods:
Exactly what does this do? Shouldn't AE2 already be able to access the arc furnace due to it using the standardized IO?

R-Kaenbyou:
> Exactly what does this do? Shouldn't AE2 already be able to access the arc furnace due to it using the standardized IO?

Due to stacksize limits in canInsertItem(), item injection from ME storage bus would be blocked. That’s because AE2(and pretty much every other logistics mod) only checks for max stacksize and inventory stacksize limit when splitting input itemstack into slots, but doesn’t respect Arc Furnace’s stacksize limits based on recipe output and upgrade level. This PR overrides default handling with a custom one, which bypasses canInsertItem checks but is also aware of Arc Furnace’s own stacksize restrictions when splitting input stacks into slots. The shift-clicking improvement uses the same item input logic, and will no longer cause all empty slots getting filled with one item first. As for collectRequested and getAvailableItemFromSlot, they are required for the MEInventory class.

ps: Huge thanks to @abel1502 , if I haven’t come across his MSU handler code, there’s no telling how long am I going to stumble before figuring out the right way to implement this.

abel1502:
I don't think this is a good way to handle this. The mod's own automation has the same problem when inserting into an arc furnace (though correct me if I'm wrong), but it's unaffected by this fix. I believe a better way would be to simply reset arc furnace automated insertion to the same mode of operation as regular machines.

Edit: just noticed it's already merged. Oh well. Not a big deal, but I kinda still stand by my point. Unless it does already work with inserters and vacuum pipes and I'm just stupid.

HbmMods:
NTM inserters will default to a stacksize of 1 if a full stack can't be inserted, so despite being slow, it should still work. Pneumatic tubes don't have this, although I could simply add this as fallback behavior, add a hardcoded exception, or add a toggle for limiting the max transferred stacksize.

R-Kaenbyou:
> I don't think this is a good way to handle this. The mod's own automation has the same problem when inserting into an arc furnace (though correct me if I'm wrong), but it's unaffected by this fix. I believe a better way would be to simply reset arc furnace automated insertion to the same mode of operation as regular machines.
> 
> Edit: just noticed it's already merged. Oh well. Not a big deal, but I kinda still stand by my point. Unless it does already work with inserters and vacuum pipes and I'm just stupid.

I am pessimistic on that… It is very difficult to move back to the regular automation mode, at least under current game mechanics. Because the same slot has to hold both input and result, there has to be some sort of stacksize limit for the input. For now canInsertItem() holds this check, but it was never intended to work with stacksize requirements, just item type. 

Obviously the input must be separated from the result in order to solve this conflict. The simplest way to handle this that I could think of is to clone (input - processed) into a list(or virtual inventory) and then overwrite the slots with the result, and clone back as the slots get emptied.

So either this, or manually adapt to ME storage buses, pneumatic tubes/conveyor inserters, etc. And I chose the more conservative latter to avoid potentially destructive reworks and bugs that could come with it.

PR #2440, opened 2025-09-19
This was an animated terra drill PR but the drill is getting axed so this is now just a bugfix and cleanup PR yay

MellowArpeggiation:
At the very least it cleans up animation handling for non-gun items (chainsaw, boltgun, crucible (sword)), and makes the system use generic enums so you can define as many different animations as you want! (up to the `short` limit, at least)

Also:
* Removes unused chunk allocation in NBT Structures
* Electrolyser metal outputs now show empty correctly on clients
* Fix entity spawn egg registration, preventing permanent conflicts with mods that do it incorrectly and no longer treating our mobs as vanilla (which is an issue, apparently? I'm not sure exactly why - go bug Legacy Modding about it)
    * note that this will cause existing mobs to despawn (other non-mob entities should be unaffected), so players will lose their precious ducks (oh no)
* Compact compressor now shows in NEI for compression recipes
* Mining drills now no longer skip over depth rock, you'll have to clear it with explosives before you can scoop up the goodies underneath


HbmMods:
grafik

PR #2441, opened 2025-09-19
Crane Structure

Lazzzycatwastaken:
Adds crane stucture, has some good weapon related loot aswell as some materials. Has some mobs with decent guns and ammo aswell as mines and traps, part of the crane crumbles as you climb up.

*i hope it clean merges because ive HAD it with shithub*

image
image


brandyyn:
ts hard

Lazzzycatwastaken:
“Type soul hard” why is he lying

PR #2448, opened 2025-09-22
Show current tool area ability next to crosshair

MellowArpeggiation:
No longer do you have to constantly double check that you don't have AoE active before mining near setups!


HbmMods:
verycat-cat-eating-chips

Issue #1781, opened 2024-12-04
Suggestion: Make hazardous gas spawns on block break modular


TeslaP09:
Hey, mod pack maker here! I noticed while mining that coal and lignite ores spawn hazardous coal dust and while digging a bit through the code, i found that that this coal gas only gets spawned on breaking 3 hard coded blocks: coal ore, coal blocks and lignite ore... Maybe this can be made more modular through config file implementation, because there is other mods, especially geology ones, that add different types of coal ore, and it would be great to be able to make them spawn coal gas too, without having to write a whole mod importing the hbm stuff... the same of course applies to other types of spawned gas too, like radon gas, you get the gist:>

TeslaP09:
The same thing goes with things like lead poisoning from items:> I‘m not a coder or anything, but generic lead poisoning effects from any item in your inventory that has an ore dict tag for lead would be amazing!!

Edit: If I may add another suggestion, it would be hella funny if the coal dust actually triggered an explosion when there‘s a player inside the dust cloud holding a specific item, like a torch or a lava bucket or if there‘s a torch placed inside the dust cloud:> Anyway, keep up the great work, this mod is amazing!!

Ashen-sysadmin:
1.7 doesn't have lead poisoning IIRC?

Issue #1892, opened 2025-01-25
Support for adding Radiation and Effects to non-HBM items*


LunatiK-ExpiX:
*- If this is already implemented, then provide better documentation of it.
So I been working on a pack that tries to unify HBM and GT5u as best as I can, and I been wanting to give GT5u's radioactive items the radiation system from HBM, so that it doesn't use the very poor system found in GT5u. However I haven't been able to figure out if you can do that, or if it is possible. So this issue I created to see if it can be added, if it wasn't already.

LunatiK-ExpiX:
I now have a few ideas on how it can be implemented
One Option is to go with the ProjectE style of adding commands to make it possible to do it ingame, while also making it possible to reload, but that might require more work. 
Another way is to make it a config file, just like all the other config files.
Lastly, there is via CraftTweaker, which might be too outdated for this.

PR #2466, opened 2025-09-30
Make it possible to add more things that produce hazards

Superslammer:
This makes it possible to add other items from other mods to the list of items that gives the "black lung" effect. It also makes it possible to make other blocks than coal and lignite to generate "coal puffs".

This is done by adding a new configuration items in the 16_pollution tab in hbm.cfg, the default values looks like this:
```
# Items that cause the coal dust hazard to apply.
S:POL_09_coalDustHazards <
    hbm:item.powder_coal:0
    hbm:item.powder_lignite:0
 >

# Items that cause the coal hazard to apply, but with a smaller effect than normal coal dust hazard items.
S:POL_10_tinyCoalDustHarzards <
    hbm:item.powder_coal_tiny:0
    hbm:item.powder_lignite_tiny:0
 >

# Blocks that can cause a puff of coal dust to appear.
S:POL_11_coalDustBlocks <
    hbm:tile.ore_lignite:0
    minecraft:coal_ore:0
 >
```
To add new items or blocks you would use this template:
\<modID\>:\<item/blockID\>:\<metadataNr\>

Please let me know if this is the right place to add this and if any changes needs to be made.
This is also created with issue #1781 in mind.

### DIFF
--- a/src/main/java/com/hbm/commands/CommandReloadRecipes.java
+++ b/src/main/java/com/hbm/commands/CommandReloadRecipes.java
@@ -1,6 +1,7 @@
 package com.hbm.commands;
 
 import com.hbm.config.ItemPoolConfigJSON;
+import com.hbm.hazard.HazardRegistry;
 import com.hbm.inventory.FluidContainerRegistry;
 import com.hbm.inventory.fluid.Fluids;
 import com.hbm.inventory.recipes.loader.SerializableRecipe;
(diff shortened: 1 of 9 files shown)


Superslammer:
Just noticed issue #1892, this PR could be part of an implementation to it, I could even expand this PR to include the other hazards if needed.

HbmMods:
Yeah it would probably be better if the system was applicable to all hazards, not just coal dust. Also maybe as a config separate from the standard forge ones, so it could benefit from `/ntmreload`, just like armor and recipe configs (for instant feedback on whether the item names are configured correctly instead of needing to restart the game every time).

David-WXYZ:
I would love to see this

WolfEclipses:
I also support this, helps with my current project too :D

PR #2473, opened 2025-10-06
common worldgen cascade fixes (improved worldgen performance)

MellowArpeggiation:
Meteorites and flowers (hemp, foxglove, etc) were frequently causing worldgen cascades (verified with ArchaicFix logging). To verify that this fix is valid, I set the spawn frequency of everything affected by the change to 1 per chunk and flew around with murky wings.

Lazzzycat this might help with your "1.7.10 worldgen sucks ball ass" issues a fair bit, and after we remove the remainder of cascades (regular, sand, bedrock oil bubbles, and the remaining old structures), worldgen performance should be considerably closer to vanilla 1.7.10


WolfEclipses:
Performance fixes for the win

PR #2475, opened 2025-10-07
Remove all remaining non-structure worldgen cascades

MellowArpeggiation:
Reimplements all oil bubbles using `MapGenBase`, including oily sand and bedrock oil.
Sellafield craters are also reimplemented to remove cascades, with the radiation generating blocks changed to slaked sellafite.


HbmMods:
grafik

scrungus

PR #2489, opened 2025-10-21
More powered lights.

DTK-MrRaccoon:
Added 3 of the existing lights but made them powered lights along wireless power supply for them with range of 3x3 chunks.


HbmMods:
but why

DTK-MrRaccoon:
> but why

replacement for electrical age lights it works in same way too and I did keep the no powered ones so if people don't want to use they don't need to

HbmMods:
they're just the lights we already have but less convenient, use double the block IDs even though this could have been handled via metadata on the existing lights, and idk why we need to scan 900k blocks every second for the weird wireless energy block thing

DTK-MrRaccoon:
> they're just the lights we already have but less convenient, use double the block IDs even though this could have been handled via metadata on the existing lights, and idk why we need to scan 900k blocks every second for the weird wireless energy block thing

yeah I know they are same but I like the ones we have but its easy to turn off with out running redstone every ware and I had no idea how to do that stuff I did this first time

PR #2504, opened 2025-10-27
Boxcables!

MellowArpeggiation:
Ever wanted thicker power cables? Or to run your power lines alongside your fluid ducts in an aesthetically pleasing manner? Want to make your base look more Greggy? Introducing boxcables!

Also included:
* Fix for PA dipoles applying maximum speed penalty when a particle isn't turning
* OC support for PA dipoles


TheVitya2127:
The cable texture looks strange, but it's definitely cool. Now we can confuse them up with cables in addition to pipes. Perfect!

gmhs022:
Oooooh hell yeah, these would also fit nicely with large power plants, use the giant ones for the direct output and step down to smaller ones with deco transformers

PR #2505, opened 2025-10-27
I've lost count of the number of times I've done this

MellowArpeggiation:
exhaust ducts crash the game


HbmMods:
ough why did you mush the cable renderer into the same class as boxducts

PR #2513, opened 2025-10-31
fix crash in Highlands mod due to it using a deprecated biome hook

MellowArpeggiation:
The [Highlands](https://www.curseforge.com/minecraft/mc-mods/highlands) mod crashes with the new cascadeless bedrock oil generation due to it using a deprecated forge hook that doesn't include block metas; this PR fixes that issue.

Bonus:
* Update foundry channels to nodespace, which should make large channels perform much better
* Arc furnace now continues to shift items from input slots while the lid is closing, so it can now fill with more than just one stack if inputting while it is idle


HbmMods:
grafik

turns out golden hammers are kinda cool, actually

PR #2537, opened 2025-11-16
Fix for floating superflat structures

Lazzzycatwastaken:
Fixes flying structures in superflat (mainly dish and lab) 

sludge


src/main/java/com/hbm/world/gen/nbt/NBTStructure.java
```diff
@@ -546,6 +550,10 @@ protected boolean build(World world, JigsawPiece piece, StructureBoundingBox tot
 					Block block = transformBlock(state.definition, piece.blockTable, world.rand);
 					int meta = transformMeta(state.definition, piece.blockTable, coordBaseMode);
 
+					if(ry < 0 || ry >= world.getHeight()) continue;
```
MellowArpeggiation: why not just make this ry < 1 and remove the `getBlock`?

PR #2549, opened 2025-11-26
Add ferrouranium block

esca111:
adds the ferrouranium block allowing it to be crafted/made in a crucible basin, also slightly changes the texture to look slightly more shiny to try to match the ingot more


HbmMods:
we're not doing any more material blocks until the whole metadata situation is sorted out

PR #2551, opened 2025-11-27
Allow machines based on ModularMachineBase (and torus reactor) to do multiple times of recipe in one tick

Creeper-banner:
Introduces a progress of pre-calculating how many times the current recipe can be done at the beginning of ticking machines. 
Behaves as what it is now if the recipe's duration is above 1 tick (as the return value would be 1), but if the recipe's duration gets below 1 tick then the modified code allows recipes to be done multiple times (i.e. the return value) , which is currently not possible.


MellowArpeggiation:
why tho

MellowArpeggiation:
soz I probably should elaborate:
why would you want these sub-ticks when you could instead just increase the input+output yields?
instead of ticking twice with 1,000mB, why not just tick once with 2,000mB?

HbmMods:
this feels horribly over-engineered for something that could be done in like two lines

Creeper-banner:
> this feels horribly over-engineered for something that could be done in like two lines

Maybe so, and yes it's a large but not very elegant work, but maybe pre-calculation could be better than sub-ticks or something?

PR #2563, opened 2025-12-01
Structure Block!

MellowArpeggiation:
A bunch of features ported back to NTM from NBTStructureLib, including:
* Structure Saving/Loading Block!
* Required jigsaw pieces - a structure will continue generating until at least one of these jigsaw pieces generates
* Jigsaw piece spawn limit - a structure will stop spawning this piece after the specified count is reached
* Platform generation - a `BlockSelector` can be defined to fill in the space underneath a structure, preventing floating
* Biome based `BlockSelector` - can be used for replacing blocks in a structure with the current biome's blocks, like dirt/sand
* Improved weight selection - for better performance and more consistent structure selection


HbmMods:
grafik

*yummers*

Vaern:
block selectors...... the more things change the more they stay the same............

PR #2580, opened 2025-12-09
Converted Radiation Absorbers to Metadata

WolfEclipses:
One block ID, 4 blocks, this is peak. Added languages because turns out it only worked in english...

Worked in testing, not sure if its how you want metadata done, copied as much as I could from other metadata blocks. If you like it ill start working on _big ones_ like the ingots and plates and wires and shit like that or I can not, whichever


WolfEclipses:
Just realized this likely will mess up all existing rad absorbers, not sure how to fix that as by changing to metadata and removing from the registry it inherently fucks it all up. Might not have major issue with rad absorbers as no one really uses them that much anyway, but would really mess up plates ingots ores and wires if converted in future. Not sure if you have a way you want to handle that, if you do lmk so I can follow that.

HbmMods:
For the transitional period it would be easiest to just leave the old blocks in, add the meta-based versions *on top of that* and then have some mechanism that mostly automatically turns the former into the latter. Back when I migrated the volcano ores I made it so that block random ticks would cause the deprecated ores to turn into the new variants. There's a new (barely tested) system though in `ModEventHandler`, the `onChunkLoad` method with some commented out demo code which is intended to translate deprecated blocks into their new variants when the chunk is loaded.

> Added languages because turns out it only worked in english...

`BlockEnumMulti` has a method for overriding the unlocalized name so that changes in the lang files aren't necessary.

src/main/java/com/hbm/blocks/generic/BlockAbsorber.java
```diff
@@ -1,39 +1,150 @@
...
-	float absorb = 0;
+    // Enum for tiers they are in order of meta data, 0, 1, 2, 3 for Base, Red, Green, Pink
+    public static enum EnumAbsorberTier {
+        
+        BASE("Radiation Absorber", 2.5F, "absorber"),
+        RED("Enhanced Radiation Absorber", 10F, "absorber_red"),
+        GREEN("Advanced Radiation Absorber", 100F, "absorber_green"),
+        PINK("Elite Radiation Absorber", 10000F, "absorber_pink");
+        
+        public final String displayName;
```
HbmMods: Don't put display names serverside, also these are seemingly not even used at all.

WolfEclipses:
Might work, forgot to test if the block is still classified correctly from past versions after I readded it, the onchuckload worked in testing, but I didn't commit it because I didn't test in on multiplayer. Will finish that in like 3 hours when I come back to this

WolfEclipses:
Works now, had exams and shit, L honestly, uh, works converting old blocks into new, old blocks are still in game marked with (LEGACY), works in testing

PR #2587, opened 2025-12-14
Make them protected so I can subclass tanks.

Ashen-sysadmin:
Please, just fix access to this stuff for subclasses. I know it doesn't affect how the mod functions as is, but it would make my downstream work so much less painful.


HbmMods:
bruv

Ashen-sysadmin:
Goddammit.

PR #2605, opened 2025-12-26
Fluid valve with counter for simpler automation

Toshayo:
A controllable valve that counts fluid that pass through. Usable for automation with RoR or OpenComputers.


HbmMods:
What exactly is the use-case for this? And why does it need a built-in switch?

Toshayo:
I'd like to automate fluid transport through a entity. To do that I'll have to trigger events when fluids reach a certain amount.
This block allows optimizing the setup: valve > tank with counting torch > valve, which will open first valve, fill the tank, close the first valve, add the amount to the final amount, open second valve, wait until the tank is empty

hacker-on-steroids:
can't get this thing to work
don't see it in the wiki or the ingame F1 helper
while i'm not fluent in java, it seems to require opencomputers or redstone-over-radio, and can't be operated manually?

Toshayo:
> can't get this thing to work don't see it in the wiki or the ingame F1 helper while i'm not fluent in java, it seems to require opencomputers or redstone-over-radio, and can't be operated manually?

You wouldn't need to operate it manually, it's designed for automation. The idea was to trigger other blocks based on fluid flow. So yes, you need redstone-over-radio or opencomputers.

**Redstone-over-radio:**
- Values: 
  - value (current counter value)
  - state (open/close -> 1/0)
- Functions:
  - reset (drops counter value to 0)
  - setState (accepts argument, 1/0 to open or to close)

**OpenComputers:**
- Component name: ntm_fluid_counter_valve
- Functions:
  - getFluid -> name of current fluid
  - getCounter -> value of counter
  - resetCounter -> drop value of counter to 0
  - getState -> gets state of the block (1/0: open/close)
  - setState -> argument 1/0 to open or close

Hope it helps :)

PR #2607, opened 2025-12-27
Rad Absorbers Metadata (Done now)

WolfEclipses:
Converts Radiation Absorbers to use metadata instead of separate blocks.

Automatically updates the block to the new one when it is loaded in a chunk.

More details in previous PR


src/main/java/com/hbm/items/block/ItemBlockAbsorber.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.hbm.blocks.generic.BlockAbsorber;
+import com.hbm.blocks.generic.BlockAbsorber.EnumAbsorberTier;
+
+import net.minecraft.block.Block;
+import net.minecraft.item.ItemBlock;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.StatCollector;
+
+public class ItemBlockAbsorber extends ItemBlock {
```
HbmMods: doubt a new `ItemBlock` is necessary since `ItemBlockBase` already handles all multi enum things, it's used implicitly by the register function that's commonly seen at the bottom of `ModBlocks` that only has a single param

src/main/java/com/hbm/blocks/generic/BlockAbsorberOLD.java
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	@Override
+	public int tickRate(World world) {
+
+		return 10;
+	}
+
+	@Override
+	public void updateTick(World world, int x, int y, int z, Random rand) {
```
HbmMods: given you're using the new migration system, i doubt it's even necessary to keep the functionality around at all (only edge case is legacy absorbers in item form, but those can be handled via item update)

HbmMods (changes requested):

src/main/java/com/hbm/items/block/ItemBlockAbsorber.java
WolfEclipses: Fixed

src/main/java/com/hbm/blocks/generic/BlockAbsorberOLD.java
WolfEclipses: removed functionality from them

WolfEclipses:
Fixed requested changes, worked in testing, I'm gonna do deco blocks (last I checked they were still normal ids not metadata) after I finish some more QMAW pages

HbmMods:
> I'm gonna do deco blocks

you should probably wait with that, given those aren't that many IDs yet are used basically everywhere in structures, we still need a proper migration tool for that.

WolfEclipses:
Alright. Fixed merge conflict, I hate whatever is going on with the craftingmanager file overwriting itself every change. Everything works again :P

HbmMods:
grafik

WolfEclipses:
Oh Great Bob, I have fixed the merge conflict, I agree, oughhhhh

HbmMods:
are you 100% sure you didn't change the file's encoding because why else would it show the entire thing as a merge conflict

WolfEclipses:
I don't think so, I wouldn't know how to check if I did, I've just been accepting your file and adding the necessary lines to make mine work

HbmMods:
just discard the changes on the crafting manager entirely, i can just add those myself

WolfEclipses:
Alright

WolfEclipses:
Fuck GitHub, I can fix it when I get home, can’t do this on my phone

WolfEclipses:
@HbmMods ok its removed now, https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/pull/2607/commits/a8b046f46bd3e0d9303317993e92b85631f2f46f contains the lines I added, the other commits with it wiped the file when I did checkout. The issue was with which computer I did it on, I modified that file on a different computer than the other ones

PR #2624, opened 2026-01-08
Better electric FSB hud bars

ranch21:
makes it possible to have more than one bar per slot, for example if you have a suit with a jetpack module it will display both the suit energy and the jetpack fuel.

if your suit does not have a jetpack module this visually changes almost nothing.

more bars can be added easily by adding to the getBars function.

suit + jetpack module
image

only jetpack or incomplete suit with jetpack module
image

theoretical >2 bars
image


HbmMods:
grafik

PR #2639, opened 2026-01-10
calculator history

ranch21:
added a history of recent operations
down arrow / up arrow to choose, enter to select

also made it render without the use of a texture

image


WolfEclipses:
This is absolutely beautiful

PR #2641, opened 2026-01-11
improved pneumatics throughput

sunryze-git:
pneumatic systems are notoriously slow when there are a lot of outputs. i did a lot of rework to the pneumatic network code which involved these changes:

- round robin now iterates through receivers until a successful transfer occurs (will bail out if we end up failed on all receivers)
- random now doesn't sort based on distance, as it doesn't need to, and shuffles the receiver list, and does the same attempts based system to ensure that each send operation tries all possible outcomes.

those initial changes made the most difference for throughput. it brought it down from multiple seconds per stack sent (in the event of 16 receivers from 1 crate), to the full speed throughput of 4 stacks a second.

some other changes i made:

- the send operation will only begin to send if the source inventory actually contains items that can be sent, which reduces CPU usage on idle networks
- validating receivers on every send call seemed redundant to me, which removed an O(N) loop. this should be done whenever update is called which im "assuming" is periodically
- invalid links are handled on the fly if that ever happens.

these changes would make pneumatic actually able to withstand complex network sizes, in the case where its not simply A-B.

https://github.com/user-attachments/assets/353d515a-216d-4335-9c7a-b06744c988e5

### DIFF
--- a/src/main/java/com/hbm/uninos/networkproviders/PneumaticNetwork.java
+++ b/src/main/java/com/hbm/uninos/networkproviders/PneumaticNetwork.java
@@ -2,6 +2,7 @@ package com.hbm.uninos.networkproviders;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


bisousbisous:
HOLLLLYYYY KINO

HbmMods:
Could you please not include mass formatting changes, it makes it hard to actually read what really changed and it breaks the very deliberate formatting that I use.

HbmMods:
image

PR #2648, opened 2026-01-13
Visible third-person muzzle flashes from other players +self, skeleguns

Vaern:
CoC:
 - Added MuzzleFlashPackets and tweaked ItemRenderWeaponBase + subclasses
   - Now, any player can see the muzzle flashes of any other entity's SEDNA gun
     - Default range of 100 blocks for packets - if needed, specific guns can have a larger range
   - Works for F5 third-person (thanks, GL_CULL_FACE!)
   - Akimbos will technically always "fire" both guns to other players, due to technical limitations (the hashmap uses entities, and is gun-agnostic) but F5 works fine!
 - Support for changing non-player entities' animation states added to ItemGunBaseNT
   - No actual animations, but allows the orchestra system to work
   - Skeleguns will not only show muzzle flashes, but play reload sounds, eject spent casings, etc.
 - Tested on SP, LAN, and dedicated MP for each affected gun + skeleguns with two clients through the IDE with no crashes/errors, so hopefully server-safe!


HbmMods:
grafik

PR #2674, opened 2026-01-24
OPTIMIZATION!!!

BallOfEnergy1:
Introduces a lot of new optimizations, especially focusing on network transactions (again).

This implements a new packet system which includes tons of optimizations:
Machines out of range of any players will no longer compile and attempt to send packets.
Packet `ByteBuf`s are no longer directly compared, instead using hashcode comparisons for increased performance.
Machines will not send data packets if the data being sent has not changed (opt-in system).
Various packet serialization optimizations (caching and such).

Alongside all these network optimizations, there's also various optimizations made to other systems, including:
Optimized conveyor crane redstone/inventory/packet systems (big thanks to Mellow!)
Also optimized many other machines that read redstone values.
Optimized nodespace net iteration (HashSets are slow).
Removing expensive unneeded reflection; replaced with _**access transformer line**_.
Fixed major slowdown in RTG blast furnaces.
Added caching in turrets for whitelist strings (to prevent constantly reading NBT).
And more!

Also added some features at popular request:
Fluid duct gauges now average out their fluid per second values (OC/ROR keep their precise values).

🌹


Lazzzycatwastaken:
gerver is so back

NoErnesto:
image-26

src/main/java/com/hbm/util/TrackerUtil.java
```diff
@@ -13,26 +13,18 @@
...
- * @author hbm
+ * <p>
+ * Now with 100% less reflection!
+ * @author hbm, BallOfEnergy01
  */
 @NotableComments
 public class TrackerUtil {
 
-	/** Grabs the tracker entry from the given entity */
-	public static EntityTrackerEntry getTrackerEntry(WorldServer world, int entityId) {
```
HbmMods, line 23: You should have left the method and just replaced the contents instead of removing it entirely. I'd rather use the conveniently named `TrackerUtil` to grab an `EntityTrackerEntry` instead of having to remember that I need a `WorldServer` then `server.getEntityTracker().trackedEntityIDs.lookup` using the entity id and then finally cast the result, all the while I'm cursing like a sailor because I only use tracker entries once every other year and I can never remember where else I used them to look at how I did it back then.

src/main/java/com/hbm/uninos/UniNodespace.java
```diff
@@ -24,8 +20,8 @@
  */
 public class UniNodespace {
 
-	public static Map<World, UniNodeWorld> worlds = new HashMap();
-	public static Set<NodeNet> activeNodeNets = new HashSet();
+	public static Map<World, UniNodeWorld> worlds = new HashMap<>();
```
HbmMods, line 23: can you stop doing that

BallOfEnergy1: my bad, my stupid IDE yells at me for `raw usage of parametrized class`

MellowArpeggiation: fixing compiler hinting to remove warnings is hardly a bad thing.

src/main/java/com/hbm/blocks/network/CraneInserter.java
```diff
@@ -188,6 +188,12 @@ public void onPackageEnter(World world, int x, int y, int z, ForgeDirection dir,
 		}
 	}
 
+	@Override
+	public void onNeighborBlockChange(World world, int x, int y, int z, Block block) {
+		TileEntityCraneInserter crane = (TileEntityCraneInserter) world.getTileEntity(x, y, z);
+		crane.isIndirectlyPowered = world.isBlockIndirectlyGettingPowered(x, y, z);
+	}
+
```
MellowArpeggiation, line 196: I now realise it should also update this flag on block placement too

PR #2692, opened 2026-02-05
Update Simplified Chinese Localization

Creeper-banner:
Only contains some small fixes in zh_CN.lang.
Maybe I would work on QMAW if I have enough time...?


TojuKrake:
can you explain what QMAW is ?
....whatever it is if you have time do it i'm a big fan of the NTM.

PR #2702, opened 2026-02-13
added third pylon with transformer but connection to single wires

DTK-MrRaccoon:
its medium pylon but with connection to single pylons too


HbmMods:
nah

PR #2712, opened 2026-02-19
HFRWavefrontObject improve

FOlkvangrField:
Improved HFRWavefrontObject. Now it can load obj models that contain both triangular and quadrilateral faces after being optimized for facet reduction.

### DIFF
--- a/src/main/java/com/hbm/render/loader/HFRWavefrontObject.java
+++ b/src/main/java/com/hbm/render/loader/HFRWavefrontObject.java
@@ -26,11 +26,11 @@ import net.minecraftforge.client.model.obj.TextureCoordinate;
 import net.minecraftforge.client.model.obj.Vertex;
 
 public class HFRWavefrontObject implements IModelCustomNamed {
-	
+
 	/** For resource reloading */
 	public static LinkedHashSet<HFRWavefrontObject> allModels = new LinkedHashSet();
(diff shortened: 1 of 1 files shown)


HbmMods:
I'm not really sure what the point of this is, all our models are already triangulated, and triangulation is a simple setting for export in most 3D modelling programs. The actual mesh that is loaded to memory is still triangles all the same.

FOlkvangrField:
> I'm not really sure what the point of this is, all our models are already triangulated, and triangulation is a simple setting for export in most 3D modelling programs. The actual mesh that is loaded to memory is still triangles all the same.

I actually discovered the problem while testing my custom machine learning model system, which is why I submitted this pull request. It seems the issue stemmed from incorrect output settings in my object model. 
Speaking of which, was there some serious problem with my previous pull requests for the custom machine learning model and sound system that led to their closure?

HbmMods:
> It seems the issue stemmed from incorrect output settings in my object model.

It would be way easier to just export the models right, then.

> Speaking of which, was there some serious problem with my previous pull requests for the custom machine learning model and sound system that led to their closure?

That PR went over half a dozen reviews and with each version something else was broken, I think the most recent version had a clietnside check on a serverside command or something like that. For a feature that we don't really need all that urgently, I really don't feel like doing a PR review every other day.

PR #2713, opened 2026-02-19
Annihilator Fix

FOlkvangrField:
Fix the custom recipe system for Annihilator. Now we can set up the pool of the output blueprints


FOlkvangrField:
@HbmMods By the way, the Annihilator system is not the final version of the blueprint research system, right?

HbmMods:
This doesn't seem to actually fix the underlying issue, that being `writeItemStack` and `readItemStack` not supporting NBT data when it should. That's probably what broke the precision assembler recipes as well. Fixed and tested in the most recent commit.

> By the way, the Annihilator system is not the final version of the blueprint research system, right?

I can't really think of any changes the system would need. Research, in essence, just means having hard gates on some recipes which require a ton of materials to be spent, and the annihilator does just that.

PR #2719, opened 2026-02-21
Added zirnox reactor redstone support on/off

Hacker6329:
I wanted to add this feature to version 1.12.2CE, but they asked me to try adding it to version 1.7.10 first.

### DIFF
--- a/src/main/java/com/hbm/blocks/machine/ReactorZirnox.java
+++ b/src/main/java/com/hbm/blocks/machine/ReactorZirnox.java
@@ -90,6 +90,7 @@ public class ReactorZirnox extends BlockDummyable {
 
 	@Override
 	public void onNeighborBlockChange(World world, int x, int y, int z, Block neighbor) {
+		super.onNeighborBlockChange(world, x, y, z, neighbor);
 		if (world.isRemote) return;
 		int[] core = this.findCore(world, x, y, z);
 		if (core == null) return;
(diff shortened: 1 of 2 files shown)


nuhyuh:
if you could un-italian the comments thatd be amazing

Hacker6329:
Ups sorry 😅

Hacker6329:
Removed almost all comments from my code except for 2 that now are in english

src/main/java/com/hbm/blocks/machine/ReactorZirnox.java
```diff
@@ -87,4 +88,42 @@ protected void fillSpace(World world, int x, int y, int z, ForgeDirection dir, i
...
+							world.getBlockPowerInput(sx, sy, sz) > 0) {
+
+							powered = true;
+							break;
+						}
+					}
+				}
+			}
+		}
+		reactor.setRedstonePowered(powered);
```
HbmMods, line 127: Without a check on whether it was previously powered, wouldn't this just shut off the reactor on every block update?

Hacker6329: Lemme check it, when i previously replied i was just waken up.

Hacker6329:
Damn i forgot to save redstonePowered value... 🤦🏻

Hacker6329:
Ok, now it should work as expected.

PR #2738, opened 2026-02-28
Make placing crates in the world require shift

abel1502:
A QoL improvement. Essentially makes crate opening from hand take priority over placing, if it is enabled at all. Previously you'd have to look at your feet or up in the air to open a crate, which was a bit inconvenient. Not the other way round (shift forces opening, regular click places) because that would make it improssible to place a crate on an interactable block

### DIFF
--- a/src/main/java/com/hbm/items/block/ItemBlockStorageCrate.java
+++ b/src/main/java/com/hbm/items/block/ItemBlockStorageCrate.java
@@ -37,6 +37,16 @@ public class ItemBlockStorageCrate extends ItemBlockBase implements IGUIProvider
 	}
 
 	@Override
+	public boolean onItemUse(ItemStack stack, EntityPlayer player, World world, int x, int y, int z, int side, float hitX, float hitY, float hitZ) {
+		// If crates can be opened from hand, prioritize this and require sneaking to place them
+		if (ServerConfig.CRATE_OPEN_HELD.get() && !player.isSneaking()) {
+			return false;
+		}
+
+		return super.onItemUse(stack, player, world, x, y, z, side, hitX, hitY, hitZ);
+	}
+
+	@Override
 	public ItemStack onItemRightClick(ItemStack stack, World world, EntityPlayer player) {
 		if(!ServerConfig.CRATE_OPEN_HELD.get()) return stack;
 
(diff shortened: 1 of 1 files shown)


Ashen-sysadmin:
No. Crate placing is the primary way of interacting with them, and the opening in hand method has caused nothing but pain to everyone.

This change would be disruptive and not actually a major QOL improvement in most circumstances.

abel1502:
@Ashen-sysadmin I'd diaagree about what is the prinary way of interaction. I definitely place crates often, but I open them from inventory a lot more. Definitely didn't cause me any pain, and it's an invaluable time-saver when mining or exporing.

More importantly, however, if crate opening from hand is disabled in the config, this PR has no effect. So if you're playing with it off, there's no downside or disruptiveness. And if you're playing with it on, the way it is implemented currently is very inconsistent and quite irritating -- you have to, essentially, always look at your feet before opening a crate.

HbmMods:
The problem I see is that people who are either new or just don't know about this change will find themselves unable to place crates entirely for seemingly no reason.

abel1502:
@HbmMods would a tooltip help?

I suppose the opposite might be true right now: new players might not realize that crates can be opened without placing them, until they do so by accident. And figuring out crates can be placed is fairly intuitive: it's a block, and it's found already placed in most structures. Shift to place a block instead of opening a UI is also a vanilla minecraft interaction, so it should be fairly intuitive either.

Alternatively, the behavior could be inverted: default is place-only, hold some key down to open. But then it can't be shift, because otherwise that would prevent anyone from placing a crate on an interactable block.

WolfEclipses:
Personally I use crates both ways, either I'm mining and open it in the air so I don't lose my crate, or I'm unloading thousands of items from structure runs by placing it down. I don't see the point in changing it at all, when all the veterans are used to it and it might be weird to the new people. Very very few people would be helped by this in my opinion.

abel1502:
@WolfEclipses I think your claim amounts to that it's a low-impact change, and I definitely agree with that, it's a very minor thing. But I think it would be helpful as it saves new players the hassle of picking up a new habit of opening the crates right. (And again, as I mentioned previously, shift-placing feels completely natural because it's essentially already a mechanic in vanilla. This is not a theoretical claim, I am playing with this change since 3 days ago and I didn't need any mental readjustments). So, while it's not a big thing, it feels like an appropriate polishing touch to me.

RedstoneParkour:
Could this be an /ntmclient option? If yes then make it default to off and everyone is (hopefully) happy

PR #2743, opened 2026-03-03
The сurium and сalifornium program - My brain  meltdown in the middle of the night

Technomirfuture132:
The result of my many hours of work, I will say right away that I am not a programmer, but I tried to do my best to graduate from this issue. And here's my suggestion - add more fuel cells from the minor actinide group to the game.

I suggest adding **_transplutonium elements_** as nuclear fuel to the mod. Yes, we have a fairly large selection, there is a shсrabidium fuel in the game, but at the moment the mod is actively developing, and the americium branch remains a dead end in this area, at least I would like to see several more special or experimental types of fuels based on americium, curium, berkelium, californium and einsteinium in the game. After **_plutonium-241_**, fissile isotopes of americium - am-241, 242 - will be produced in a reactor such as RBMK, Watz or PWR. After processing the spent americium-based fuel, valuable curium isotopes can be separated in SILEX or PUREX, which in turn can be divided and used in mixed curium fuels (based on americium-curium, plutonium-curium, uranium-curium) or added as salts to a special coolant. In any case, progress in the game of increasing the atomic weight above plutonium will be much slower, since the volume of transplutonium elements produced will be smaller compared to the same number of transuranic elements.

Another application of americium in the game that I see is its use for the reproduction and production of curium in reactors such as RBMK, PWR and an experimental reactor with powerful neutron fluxes. Americium has many isotopes, **_americium 241 and 242_** are currently represented in the game, but in reality there are many more of them, for example, the same americium-241 isotope can be transformed in a reactor first into _**americium-242**_ by capturing a neutron, then capturing a neutron again, it will turn into **_americium-243_**, and that into its own queue at _**Americium-244.**_ Of course, in real life everything is much more complicated and it is also necessary to take into account the neutron capture cross-section, which affects the probability of the nucleus capturing a neutron and turning into the next isotope. **_Americium-244_** can decay to form curium-244 by emitting an electron, while the decay of its other isotopes mainly results in fragments of different atomic masses, that is, a lot of nuclear waste.

The new fuel should have a shorter life span, but also generate more heat. It is necessary to forget about highly enriched fuel, it will be extremely unstable - it is best to use it in the above-described layout schemes, creating more variety in nuclear transformations. Berkelium-based fuel, a by-product of curium-americium fuel burnout, will be even more radioactive and rare compared to curium, but very little of it will be needed as part of the new nuclear fuel. In addition, berkelium can be obtained artificially, on a cyclotron. As for the subsequent nuclear transplutonium isotopes - **_californium and einsteinium_** - they will be used as experimental fuel cells in extremely small quantities, perhaps 1 nugget per 8 slots of mixed nuclear fuel.All these elements can also be obtained by laser separation in the visible light spectrum of nuclear fallout from nuclear explosions, reducing the likelihood of dust escape. All these elements can also be obtained by laser separation in the visible light spectrum of nuclear fallout from nuclear explosions, reducing the likelihood of dust escape, or artificially obtained in an undivided form in the form of a powder on a cyclotron.

Obtaining nuclear fuel based on curium, berkelium, californium and einsteinium will also require adding new types of liquid to the game - if possible, liquid analogues of fluorine-based transplutonium salt can be created from fluorite or chlorine, such a eutectic mixture will be a more effective neutron modifier compared to thorium salt in PWR, the rest can be added to your vision future modifications.

And of course, it is possible to add more variety for weapons-grade applications of transplutonium fuel, as **_curium-245 is an excellent fissile material for creating bombs with a much higher fission coefficient than uranium-235 and plutonium-239_**. This means that one nugget of curium-245 will be equivalent to a whole ingot of plutonium-239 or about 3-4 ingots of uranium-235 by the increment factor in a custom bomb. **_Californium-252_** is the strongest source of neutrons, it can be used as a component for a dirty bomb. And there are also a lot of applications for these elements, such as creating RTG cells with extremely high efficiency and self-charging batteries producing 10k HE/t and much more.


RayzerHan:
You did a great job, but I'm wondering if you used a translator to translate the text into Russian, or are you a native Russian speaker?

70000hp:
what

70000hp:
have you read CONTRIBUTING.md before submitting this PR

HbmMods:
But why, though?

The reason the fuel chain ends with Americium is because returns are diminishing with every step, so actually getting Americium, even with ZFBs, isn't really worth it. This adds 15 new materials, which are awful to get, and are mechanically no different from any other fuel we already have. Any HE fuel, Schrabidium, or even Balefire already do the "generate more heat" part while being way easier to obtain. I did consider adding Californium a while back, but with the condition that it would have a unique mechanic that would actually make it worth using (high durability + temperature coefficient, making it both powerful and considerably safer), and even then I hesitated because realistically, would anybody even use it?

By the way, your PR would break the mod anyway, the changes in the `Fluids` class contain a copy of a ton of fields which wouldn't even compile, and the addition of the new waste item enums in the middle of the enum list would ID shift existing schrabidium waste items.

> The result of my many hours of work

As `CONTRIBUTING.md` states, you might wanna ask before spending this much time on a PR that's guaranteed to not be merged.

Technomirfuture132:
NTM-mods, it's a pity, it would be quite interesting, because the chain ends in americium, which makes the mod somewhat incomplete. Schrabidium and australium, although they are fictional elements, in reality they could be elements of the "island of stability", but before that you have to go through a complete chain of nuclear transformations. Do you know the mod for the Nuclear Craft? If not, then I liked the implementation of the nuclear transformation system available in this mod in version 1.12.2. In my opinion, schrabidium and australium are the elements of the final part of the game, there are many more elements that can be added before it, the existing RBMK rod system can be redone. 

Mark old rods as "legacy", new rods will have autogen depending on the type specified in the itemRBMKrod file , the color of the rod is also indicated using rgb numbering, the type of reactivity and depletion function is also indicated in the same file. As for reactivity, the new fuels are mainly mixtures of uranium, plutonium, and americium fuels with curium, berkelium, and californium, which have high reactivity and a low xenon index by default. Although californium is an extremely unstable element, it can be used in neutron source rods and in a research reactor. You say that you wanted to add californium to the game, but since no one was interested in using it, you abandoned this idea. However, an alloy of californium and lead, for example, will make it less radioactive and more refractory. It can also be used as a melt in new types of liquid salt nuclear fuel reactors. I don't see the point in using zfb fuel in the Zirnox reactor, since it doesn't provide anything new at all. Instead, you could use the fast neutron function to multiply and reproduce new types of fuel in the reactor and burn out nuclear waste. Since the reactor mainly fuses fuel, but also captures neutrons, when the balance shifts to capture neutrons, more fissile fuel is produced from non-fissile fuel. In the mod itself, I liked the reasim concept, I want you to add a neutron flux in all directions to the modification for new and old fuel cells by default. 

It is also possible to rethink the radiation itself in modification. For example, if you add different types of beta, alpha, gamma, and neutron radiation modifiers to the existing radioactive elements in the mod, it will add some variety to the modification. There are various categories of hazards in mod, such as asbestos, coal dust, and so on, as well as some filters for gas masks with varying degrees of protection against these types of hazards. 

Could you add new categories of different types of radiation to the mod? Low-level armor, which is easier to create at the beginning of the game, will shield beta and alpha radiation well, can save you from some neutron radiation, and only heavy power armor, lead liquidator armor, or armor with lead overlays can save you from more penetrating gamma radiation, but not completely. For each type of armor, it is necessary to provide degrees of protection against 4 types of radiation. Schrabidium and balefire must create powerful gamma radiation, and work with them is possible only by automating fuel loading, unloading and reprocessing.

Of course, you have your own vision for the future of this modification, and I am grateful that the mod is actively developing, some branches of the mod have become full-fledged and independent branches from the main modification on version 1.7.10. However, I think my advice will be useful for developing the mod in the future.

HbmMods:
i thought about writing a whole in-depth response but like, what's the point. you're just listing a ton of things that would be ReAlIsTiC, telling me how "interesting" they would be, while ignoring the original question of *why* they would be interesting. i add features because they add something to the game, something of value, some purpose, i genuinely do not give two shits about crossing off some entries from a checklist of what isotopes could be fissile that we should definitely have in the mod because "they're interesting" even though they behave the exact same as the 50 other existing isotopes we have already.

> However, I think my advice will be useful for developing the mod in the future.

no, not it really won't be. you're not the first, second or fiftieth person to make suggestions like that. please just go do something else with your time.

PR #2766, opened 2026-03-09
Update console.json

Xavier-Apoapsis:
Since guthib isn't showing the changes nicely I'll specify them here.

From: "making it easy to detect changes such as a drop in reaction due to fuel depletion or hazardous reaction spikes"
    To: "making it easy to detect changes such as a drop in the reaction due to fuel depletion or a hazardous reaction spike"

From: "there is a small text field that can set the control rod extraction level of all selected control rods"
    To: "there is a small text field that can set the extraction level of all selected control rods"

From: "Finally, the small white/red gradient button allows to cycle between the compression settions"
    To: "Finally, there is a small white/red gradient button which cycles through the compression settings"

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,11 +1,10 @@
 ## Changed
-* Updated russian localization
-* Safes now properly use the inventorytweaks API, allowing slot sorting via GUI
-* The meltdown disabling dial now also affects meldown caused by breaking fuel channels while hot
-* The dots on the grpahite moderator columns are now a lot darker to tell them apart from fuel channels much more easily
+* Updated textures for some nuclear bomb GUIs
+* RBMK fuel rods too hot to be taken out manually can now be removed by players in creative mode
+* Updated the SILEX recycling recipe for MOX fuel to reflect the recent recipe change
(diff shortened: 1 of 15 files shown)


Xavier-Apoapsis:
There is a different commit of this same file which I did incorrectly (typo) but I don't see a delete or undo button to make it go away.

WolfEclipses:
There is no listed second commit, it only shows one.

Xavier-Apoapsis:
I haven't tried to merge it since it's not correct.

WolfEclipses:
IF you did not merge it then don't worry about it, it does not exist in our eyes.

Xavier-Apoapsis:
Yeah but it appears as something I could merge and I want it to go away because it's a mistake.

Xavier-Apoapsis:
Before I try this again it's clear I have a lot of learning to do. Ideally I'd like to push multiple changes in one go with each clearly commented, and as far as I know this is possible, I just need to figure out the procedure.

Xavier-Apoapsis:
Not what I meant to do but all the commits are gone now so ok.

PR #2773, opened 2026-03-10
less hideous steel tools

sten104:
no, this isn't the only thing i'm doing, i'm working on the italian localization since it has tons of grammatical errors and just straight up wrong translations (and its not up to date)


HbmMods:
yeag

PR #2788, opened 2026-03-15
well i guess i'm committing this too

sten104:
the steel sword is now consistent with every single other ntm sword, and now looks like a medieval longsword


purpl3xity:
Mmmm I liked the vanilla look better

WolfEclipses:
Why are those in the same PR

sten104:
this is my first time using github with the ide, i have no idea how it works

MellowArpeggiation:
wack

MellowArpeggiation:
> this is my first time using github with the ide, i have no idea how it works

you should make sure to separate different things into different git branches, for your own organising and to prevent this sorta thing

HbmMods:
In this case I'll probably accept all the changes anyway. The steel sword was kept small because of how low tier it is, but honestly, that's a pretty bad reason not to have it look actually cool. And the template folder is scheduled to be executed by firing squad anyway, so moving everything out will finally force me to add a recipe selector to the crucible like i should have months ago.

sten104:
glad i could be useful for once

PR #2797, opened 2026-03-19
made strand caster work better

legendarydoge30:
strand caster now casts its contents as soon as it is able to cast

i literally just changed the number of molds to cast from 9 to 1
it was a single number

i didnt bother completely removing the feature because 
1. high chance to break the thing (common theme regarding the strand caster)
2. strand caster will be reworked in the near future anyway so its just little fix until that day comes

accept if you want, it would make life a lot eaier and would stop strand caster clogging in big factories (primarily ntm:space bedrock ores)

pleaseeeee :3 :3 :3


HbmMods:
I mean I guess that works but you could reduce the entire condition down to `moldsToCast > 0` since the second half just means the same thing again, so if `moldsToCast > 0` is true then `moldsToCast >= 9 || worldObj.getWorldTime() >= lastProgressTick + 200` will always be true anyway.

legendarydoge30:
i wanted to change the code as little as possible

PR #2799, opened 2026-03-20
how do you get a cylinder unstuck from a barrel connector

sten104:
- barrel connectors 
  small visual part that pops up when you connect it to a fluid duct
  slightly downscaled the hazard diamond to make space for it


HbmMods:
grafik

PR #2812, opened 2026-03-26
Enable 2D texturing in RenderICFController

12411234:
Texture Fix for Angelice Render Mod


HbmMods:
This was already fixed in the most recent commit.

PR #2814, opened 2026-03-27
3 lines of code can make wonders

sten104:
made bullets render fullbright so they look good at night


sten104:
actually i found an issue where things like the hook and demolition charges also get fullbright, i'll be back

MellowArpeggiation:
fullbright is implemented in like 50 different ways all over the place we really gotta unify that

PR #2829, opened 2026-03-31
make RBMK panels and meters not break instantly

hacker-on-steroids:
[5% tested]
the meters are cool as fuck, but left clicking them without a pickaxe immediately breaks them without dropping
i hope i did inheritance right and this won't break something else that depends on it being unset

before

https://github.com/user-attachments/assets/281426cd-ed95-45e0-b3d4-b5b7c57aa37d

after

https://github.com/user-attachments/assets/2646771b-a7ff-45ba-8bcc-30e0dc22395e

### DIFF
--- a/src/main/java/com/hbm/blocks/machine/rbmk/RBMKMiniPanelBase.java
+++ b/src/main/java/com/hbm/blocks/machine/rbmk/RBMKMiniPanelBase.java
@@ -24,6 +24,8 @@ public class RBMKMiniPanelBase extends BlockContainer implements ISBRHUniversal 
 
 	public RBMKMiniPanelBase() {
 		super(Material.iron);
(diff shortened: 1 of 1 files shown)


legendarydoge30:
bob is aware
its being fixed
⚠️ 
homonculus

legendarydoge30:
oh this is a PR i thought it was an issue
also probably test the thing anyway even if it just changing the hardness

hacker-on-steroids:
took me a while to figure out how to work the gradle (you can't just click the .bat to get your .jar) but i did it

hacker-on-steroids:
now they just need a tooltip or something saying that you need to shift-screwdriver to configure but i'll leave that as an exercise for the reader

HbmMods:
grafik

PR #2833, opened 2026-04-05
la ricetta per l'amore

sten104:
-buffed normal steel recipe(3 iron to 1 carbon) which should make less awful ratios(now you can make a full steel ingot with 1 coal and 1 iron!) and should reduce the times you get "1 nugget 2 quanta" residues

-added an alternate steel recipe that makes 50% more steel at the cost of more carbon and flux (2 iron 1 carbon 2 flux), also makes slag as a byproduct

-both are still weaker than using the blast furnace iron ore recipes


HbmMods:
> -added an alternate steel recipe that makes 50% more steel

we are *so* not doing that

PR #2840, opened 2026-04-08
Oil should float, this one doesn't (QMAW)

WolfEclipses:
QMAW fluid update, gonna add one a day until I've finished oil up

### DIFF
--- a/src/main/resources/assets/hbm/manual/fluids/heavy_oil.json
+++ b/src/main/resources/assets/hbm/manual/fluids/heavy_oil.json
@@ -1,0 +1,15 @@
+{
+	"name": "Heavy Oil",
+	"icon": ["hbm:item.fluid_icon", 1, 12],
+	"trigger": [],
+	"title": {
+		"en_US": "Heavy Oil",
+		"ru_RU": "Тяжелая нефть",
+		"zh_CN": "重油"
+	},
+	"content": {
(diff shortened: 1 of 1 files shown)


HbmMods:
Not sure if listing all the machines that heavy oil can be processed in is all too useful (especially ones in like oil 3, which isn't relevant to people who are likely to look up heavy oil at all), while omitting things like the endpoints of heavy oil processing, like asphalt and petroil, which would seem more immediately useful in that context. Perhaps I should write a style guide for QMAW, since I want them written in a very specific way.

WolfEclipses:
> Not sure if listing all the machines that heavy oil can be processed in is all too useful (especially ones in like oil 3, which isn't relevant to people who are likely to look up heavy oil at all), while omitting things like the endpoints of heavy oil processing, like asphalt and petroil, which would seem more immediately useful in that context. Perhaps I should write a style guide for QMAW, since I want them written in a very specific way.

That would be helpful, thank you Bob

HbmMods:
[Style guide is up](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/master/src/main/resources/assets/hbm/manual/STYLEGUIDE.md), it's not terribly precise but I hope it does get the general idea across.

WolfEclipses:
Ok thank you, I'll give it a try sometime in the next day or so, lmk what I need to improve/change then 🫡

PR #2851, opened 2026-04-15
OpenComputers methods for rbmk gauges, graphs, keypads and numitrons

DTK-MrRaccoon:
Added OC compat for the 4 new components.


HbmMods:
Doesn't OC already have its own buttons and displays and whatnot? Also, when would it ever be useful to switch the function of a fixed position button?

DTK-MrRaccoon:
No OC doesn't have any thing like this simple display/input devices. It is useful for multimode panel where single panel can have multiple roles based on setting or one button could switch between pages and set the other ones to do different things.

PR #2862, opened 2026-04-24
Added more RoR functions to zirnox and gas turbine.

DTK-MrRaccoon:
more options for control and monitoring


HbmMods:
grafik

yum

PR #2872, opened 2026-04-28
Numitron Enhancements for OpenComputers

JoelBenReneHecht:
I did create a suggestion for this but as I have decent coding experience so decided to give it a shot myself.

Without OC the behaviour should be identical.

I conducted in-depth testing client- and serverside with all the functions on the first display, the second display only received sporadic testing.
[text_cases.txt](https://github.com/user-attachments/files/27162309/text_cases.txt)

Changes I made:

- Toggle individual digits: it can be selected which digits are enabled and disabled
- Toggle leading zeroes: it can be toggled if leading zeroes are added
- Display mode: there are 3 additional display modes:
  - Normal
  - Integer: The full integer is displayed. Values above 9999999 are displayed as 9999999. Values below -999999 as -999999
  - Signed Hexadecimal: Values above 0x7FFFFFF are displayed as 7FFFFFF. Values below -134217728 (0xF...F8000000) as 8000000
  - Unsigned Hexadecimal: Values above 0xFFFFFFF are displayed as FFFFFFF. Values below 0 are displayed as -0
- Negative numbers were not displayed properly
  - leading zeroes are now added between the '-' and the number
  - Negative tripple-digit numbers with a magnitude suffix lead to a 8-character string to be returned from `getShortNumber()`. This function was reworked to have one less decimal place in such case, the behaviour should otherwise be unchanged


JoelBenReneHecht:
As I also had the idea of being able to display strings on the numitron I decided to make a texture for it based upon a 14-segment display. However before starting to implement that I would like to know if that has any chance of being accepted as that would likely take a few hours which I wouldn't want to waste.

numitron_14_segment

BallOfEnergy1:
With the OC API conventions, I would recommend throwing an IndexOutOfBoundsException upon an invalid index rather than the soft-exception approach of returning false and the error. Generally soft-exceptions should be used for cases in which, for example, an item in a list cannot be found. Index/bounds checking in a setting function like this should probably use hard-exceptions to ensure that programs are easier to debug.

Though, on the contrary, since these are set-and-forget functions that have their return boolean dictate their success, it's completely valid to leave these as-is.

Other than that, I quite like the way this OC code looks.

HbmMods:
I think I've mentioned this somewhere else before, but doesn't OpenComputers have its own displays? I'm honestly not really a fan of adding so much stuff to the numitron code in order to make it do things it was never meant to do, and honestly doesn't really need to.

If we really *need* RoR displays with this much functionality, actual text screens with RoR value support are already planned. The aesthetic will be different, but functionally it will be able to do what you're adding to the numitrons and some more stuff on top of that.

JoelBenReneHecht:
Well OpenComputers does have it's own displays, yes, but I wanted to at least display larger integers on the numitron without them being converted (e.g. that 1600 would be displayed as 1600 instead of 1.60K).

However it is possible that I may have gotten one to many ideas whilst thinking about this.

HbmMods:
Not shortening the number is a good idea though, but one that should ideally also exist in the GUI so that proper RoR can benefit from this.

JoelBenReneHecht:
Ok so I edited the GUI now.

Shortening the number and leading zeroes can now be toggled in the Numitrons GUI.

However, as I now don't have any time left rn I wasn't yet able to properly test it. Here are the test cases I'll check this evening:
[text_cases.txt](https://github.com/user-attachments/files/27240710/text_cases.txt)

I'll report back when I tested it and fixed any found issues.

Btw about the other two features with deactivating digits and leading zeroes: are these ok to stay? Considering that there is enough space left in the GUI below the buttons, should I also add the ability to select which digits are active, or should this stay OC only? Or should I remove one or both of these features entirely?

HbmMods:
For regular RoR operation, which is all that concerns the GUI, I don't see any reason whatsoever to disable digits. For OC, I'm not sure about the use-case either, but since the values are fully custom, you could probably use it for something like disabling the center digit and have two three digit displays in the same display unit, so I suppose there's some utility there.

JoelBenReneHecht:
I finished testing now.

Apart from one bug I fixed I noticed something with the selection-boxes for text input fields in the GUI:

It seems to be there with all textboxes on other blocks aswell (so probably not something I did?): the 'box', in which I have to click to select it, is extended a bit down, so clicking a bit below the green box (even slightly below the gray frame around the text box) already selects it. Now from the way the text boxes are defined in the source code this seems obvious and it _looks_ right, so I assume that this is just how this font is and with another font the displayed text would likely occupy the entire clickable portion?

About the icons for leading zeroes and shortening the number: idk how good they actually are (they seem perfectly fine to me) so you might want to rework them.

src/main/java/com/hbm/render/tileentity/RenderRBMKNumitron.java
```diff
@@ -73,7 +104,7 @@ public void renderTileEntityAt(TileEntity te, double x, double y, double z, floa
 				else if(c == 'T') {u = 0.3; v = 0.5;}
 				else if(c == 'P') {u = 0.4; v = 0.5;}
 				else if(c == 'E') {u = 0.5; v = 0.5;} // i would love to say this sucks, but this is actually surprisingly easy to read and probably the most performant way of doing it
-				int charVal = c - '0'; // no string operations, no int parsing, no nothing, we just rawdog shit shit
+				int charVal = c - '0'; // no string operations, no int parsing, no nothing, we just rawdog this shit
```
HbmMods, line 107: thanks

PR #2898, opened 2026-05-06
[NEW] Added item "falloutball"

Woowz11:
It's like snowball, but made from fallout.

It can be thrown and infect the creatures it hits with radiation.

This is my first ever pull request, so there may be some errors in the filling.


src/main/resources/assets/hbm/lang/ru_RU.lang
```diff
@@ -2677,6 +2677,7 @@ item.explosive_lenses.desc=Сборка из 8 линз из пластично
 item.factory_core_advanced.name=Энергетический кластер усовершенствованной фабрики
 item.factory_core_titanium.name=Энергетический кластер базовой фабрики
 item.fallout.name=Куча радиоактивных осадков
+item.falloutball.name=Радижок
```
TheVitya2127, line 2680: "Радижок"? Interesting name

MellowArpeggiation:
this ain't a scene, it's an arms race

oh you said BALL

Troodontid:
Isn't this essentially the same thing as the nuclear waste pearl?

HbmMods:
Well yeah it does look like a worse version of the nuclear waste pearl, which at least had some AoE to it due to the dense radon it could spawn, and the nuclear waste pearl died in the grenade rework for being effectively useless.

src/main/java/com/hbm/entity/projectile/EntityFalloutball.java
```diff
@@ -0,0 +1,65 @@
...
+
+			if(!this.worldObj.isRemote && entity instanceof EntityLivingBase) {
+				if(entity instanceof EntityPlayer && ((EntityPlayer)entity).capabilities.isCreativeMode) return;
+				PotionEffect effect = new PotionEffect(HbmPotion.radiation.id, 10 * 60 * 20, 0);
+				effect.setCurativeItems(new ArrayList());
+				((EntityLivingBase) entity).addPotionEffect(effect);
+			}
+        }
+
+		if(this.worldObj != null && Minecraft.getMinecraft() != null && this.worldObj.isRemote) {
```
HbmMods, line 45: Yup, that's gonna crash servers. `Minecraft` is clientonly.

Woowz11: okay, i gonna fix that

Woowz11: or not worth it?

HbmMods: Well again, it seems like a worse version of a feature we already had and subsequently removed for being entirely pointless, so I don't see why we would need that.

Woowz11: for fun............... 😁😁😁😁😁

It's just that when I was playing, I was confused by the fact that the radioactive fallout looks like snow, but what falls out of it isn't something like snowballs, and I decided to make it

Woowz11: I'll fix the code anyway, and you'll still have to look and think about whether it's necessary or not. If you don't accept my "радижок", I won't do pull requirements anymore.... it's a threat!!!

Woowz11:
#2901

trollbebra:
ts wouldve been fire bro

PR #2903, opened 2026-05-08
Add mouse hover overlay for HE/RF converters

MellowArpeggiation:
gaba

also new light models from Vitya because the ones I made like 2 years ago were shit


HbmMods:
grafik

PR #2906, opened 2026-05-09
OC compat for terminal

DTK-MrRaccoon:


HbmMods:
What does `ocDisplay` even do? It's only actually used by an OC command for reading it, at which point, couldn't you just cache the previous command on the OC script instead of sending it to the terminal and then reading it from the terminal again? Also, why is it serialized and send to the client, even though nothing else uses it?

DTK-MrRaccoon:
Yeah you right, it was mostly me not wanting to touch existing code so I did make new one but I will improve it. I did over complicated it.

PR #2908, opened 2026-05-10
Heavy oil qmaw v2?

WolfEclipses:
Ok so this time I tried to follow the new style guide, doing this as a trial run for A: my new workflow of making it look right then making it in the actual software (cause faster) and B: if this is the style you want, if so then we ball


HbmMods:
yum

Issue #2838, opened 2026-04-08
Industrial turbine: Pulsed steam input increases efficiency to nearly 200%


1465114862:
mod version: X5649
I guess that is not intentional.

<img width="2560" height="1377" alt="Image" src="https://github.com/user-attachments/assets/58859bb8-b608-4ba6-9f40-40ebd083f2e2" />

How to reproduce:
As shown in the figure.
1.Place a fluid tank to provide infinite steam, and a fluid tank to suck all low-pressure steam. Connect them to industrial turbine. The steam input is control by a redstone fluid valve. Put a lever on the redstone fluid valve.
2.Use a Etho Hopper Clock to control fluid valve for steam input. I use a NOT gate to let on-pulse a little longer than off-pulse. Put 5 items in the hopper.
3.Turn the lever on. Until the spin of the industrial turbine reach 100%. Turn the lever off.
4.You will see the spin will fluctuate between 85% and 100%, and steam consumption was halved by the fluid valve. So we increases efficiency of the industrial turbine to nearly 200%.

Why this happened:
The acceleration of industrial turbine is const, for most of the time. So if you stop inputting steam at 100% spin, the spin will drop at a constant speed. And then, you input steam. Input time as the same as the time you stop. The spin will return to 100%. Let's say, the minimum spin during this process is 85%. But the steam consumption was halved. So overall, efficiency exceed 100%. Ideally, the efficiency is close to 200%.

Suggestion:
When consuming steam, make the amount it comsume proportional to the increase in the energy of the spinning object. When generate power, make the amount it generate proportional to the decrease in the energy of the spinning object. Then introduce friction.

Or at least, make the efficiency can not exceed 100%, and take the maximum efficiency when running at constant speed.

HbmMods:
I can't replicate your setup because the weird piston hopper hybrid clock keeps getting stuck, so I opted for a simple repeater clock instead.

<img width="1920" height="1145" alt="Image" src="https://github.com/user-attachments/assets/25e065f1-11e2-42b5-ae82-55bae55d4b1f" />

Taking the fluctuations into account, power produced and steam used seem to check out more or less.

> When consuming steam, make the amount it comsume proportional to the increase in the energy of the spinning object.

That's already the case, the rotor's acceleration is based on the steam consumed.

> When generate power, make the amount it generate proportional to the decrease in the energy of the spinning object.

That is *also* already the case.

1465114862:
[...]

PR #2910, opened 2026-05-11
attempted turbine efficiency fix

hacker-on-steroids:
attempt at #2838
makes turbines' power output follow an exponential curve instead of linear when disturbed

https://github.com/user-attachments/assets/f323bab0-d56c-4d15-93e1-943e85241e9f

notes:
-if spin is intended to be RPMs, it could probably be tweaked to satisfy image
-industrial turbines are 1:1:1:1, while singleblock and leviathans need 10x the turbines as you go down each steam density tier
-FLYWHEEL_MAX_ENERGY could probably be reduced for dense steam turbines so they don't take years to spool up/down
-no idea how `this.variable` and `variable` are any different
-not familiar enough with java to know if multiplication is faster than branching but i removed some `if`s anyway
-copying code from TileEntityTurbineBase.java hurts my soul but since this is an abomination anyway i'll leave it to you to copy the idea and make it to standard

### DIFF
--- a/src/main/java/com/hbm/tileentity/machine/TileEntityMachineIndustrialTurbine.java
+++ b/src/main/java/com/hbm/tileentity/machine/TileEntityMachineIndustrialTurbine.java
@@ -32,9 +32,11 @@ public class TileEntityMachineIndustrialTurbine extends TileEntityTurbineBase im
 	public float lastRotor;
 	
 	public double spin = 0;
-	public static double ACCELERATION = 1D / 400D;
+	public static double FLYWHEEL_MAX_ENERGY = 0.5e8; //aka flywheel mass
+	public long maxPower = 0;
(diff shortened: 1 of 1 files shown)


HbmMods:
How long does the spinup take on the various steam types? Because right now both spinup and spindown on every steam type takes 20s, which is the intention.

hacker-on-steroids:
well because it's exponential it technically takes infinite time to fully settle, but a rough estimate for 0%-73% is 8seconds for normal steam and 80 seconds for the dense ones

HbmMods:
Perhaps not ideal but at that point sure, whatever

hacker-on-steroids:
https://github.com/user-attachments/assets/2863b06e-6155-4f8a-8ee0-4ac530825aae

PR #2911, opened 2026-05-12
Thats not light oil

WolfEclipses:
Qmaw for naphtha? Or is it, its not light oil though


HbmMods:
How does naphtha become high performance solvent? I mean technically you can process any oil into either refgas or some heavy oil for tar, but I don't see any direct ways that would need mentioning.

WolfEclipses:
> How does naphtha become high performance solvent? I mean technically you can process any oil into either refgas or some heavy oil for tar, but I don't see any direct ways that would need mentioning.

Yeah that's fair, forgot that like everything can do that

WolfEclipses:
@HbmMods fixed

PR #2912, opened 2026-05-12
It won't stop floating

WolfEclipses:
This is light oil


HbmMods:
Light oil fractions into diesel, not gasoline. Also Light oil is fully capitalized in the second sentence even though we're supposed to use sentence case here.

WolfEclipses:
@HbmMods fixed, I am never doing this late at night again xd

PR #2914, opened 2026-05-17
Gun animation reload

FOlkvangrField:
Allow to reload specified gun animation using `/ntmreload animations <name>`, or reload all gun animations using `/ntmreload animations`. Use the `/ntmreload animations list` to display all available animation names.


HbmMods:
1. Have you tested this on servers at all?
2. Why does this need to be a `/ntmreload` sub command instead of being part of resource reloading?
3. Is it really necessary to have all this `AnimPath` stuff which effectively means that animations need to be registered twice instead of just integrating that into the `AnimationLoader`?

PR #2926, opened 2026-05-31
Configure maven publishing

MartinTheDragon:
due to some demand actually existing, i updated the buildscript to allow automated publishing to our maven repository at <https://maven.ntmr.dev>

to actually publish things on it, you first need a `maven.properties` file in the project's root directory, with the `username` and `password` fields set to the credentials i will grant you in private, then you simply run `./gradlew publish` to get a release up

i already published the release from earlier this month for you :p

if you happen to accidentally publish a release, it will have to be removed from the maven repository before a release under the same version number can be republished. for that, you'll have login to the maven repository using the credentials i will provide and hit the trash icon on the release

the buildscript now also generates an unobfuscated `dev` and an uncompiled `src` jar for convenience. i considered setting up an `api` jar as well, but the api is currently essentially unusable for inclusion in other mods and also doesn't offer anything particularly useful, so i refrained from doing that

ps: as a bonus i also excluded unnecessary dependencies from the build by making most of them non-transitive, so the thing loads faster now :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,8 @@
 # CurseForge configuration
 /curseforge.properties
 
+# Maven configuration
+/maven.properties
+
 # Changelog backup
 /changelog.bak
--- a/README.md
+++ b/README.md
@@ -24,6 +24,26 @@ For further ports, try:
 
 Simply navigate to "Releases" on the right side of the page, download links for the compiled JAR as well as the corresponding source code are under the "Assets" category below the changelog. Make sure to review all changelogs when updating!
 
+## Using in your project
+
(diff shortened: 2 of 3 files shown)


Toshayo:
Can't you pull the mod via curseforge's maven and modrinth maven?

MartinTheDragon:
those are 3rd-party, update rarely, and have required deobfuscation by anyone trying to use them, which i'm not sure is even something that stock forgegradle for 1.7.10 provides

Toshayo:
NTMR is also 3rd-party. Forgegradle handles correctly deobfuscation allowing users to make dependencies on curseforge's or modrinth's mavens. As for the updates, the buildscript already has publishing on curseforge, so your maven probably won't be updated more often.

PR #2930, opened 2026-06-02
Crane Structure

70000hp:
Additions (content side):

- Heavily reworked crane structure, features several logic block puzzles, updated loot, heavy utilization of enemy spawns
- Fans can now be switched to suck in entities using the defuser

Additions (dev side):

- Logic blocks by default will be invisible, unless a disguise is selected
- Loot blocks can now be triggered using a detonator, locked using locks, and have their settings copied via settings tool
- Structure debug mode can now be toggled ingame via /ntmserver
- Structure loading block can now load already transformed structures (as opposed to the debug version) with a redstone signal
- Enemy inaccuracy can now vary based on the gun they are using, which prevents extremely inaccurate sniper weapons, and extremely accurate shotguns
   - Enemy accuracy values can now also be changed externally to make them more or less accurate depending on context (i.e the soot mobs remain just as accurate, while logic block spawns can be more accurate if needed)


HbmMods:
i only skipped over most of the changes, if something breaks i'll just hack your limbs off with a rusty machete

70000hp:
WHAT

PR #2933, opened 2026-06-03
AUTOCAL "Desktop.browse" Linux function fix

Voxelstice:
changinging the changes:
- fixes AUTOCAL ``Desktop.browse`` not working on Linux (throws a ``java.lang.UnsupportedOperationException``)

I don't know why GitHub isn't showing the diff properly but the only thing I actually changed is the ``browse`` references to a custom function in the class, and then the actual ``browse`` function implementation is below ``public GUIScreenRadioAUTOCAL(TileEntityRadioAUTOCAL autocal) {``

should we even be doing cross-platform shit like this? i don't exactly like this solution, especially when it involves running a command. i also have a feeling this isn't broken on all distros but i am not installing every single distro to find out. the most testing i was able to do is on Debian 13 and CachyOS (which is essentially Arch Linux anyway)

### DIFF
--- a/src/main/java/com/hbm/inventory/gui/GUIScreenRadioAUTOCAL.java
+++ b/src/main/java/com/hbm/inventory/gui/GUIScreenRadioAUTOCAL.java
@@ -9,8 +9,8 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.net.URI;
 import java.util.Arrays;
-import java.util.Objects;
 
+import org.lwjgl.Sys;
 import org.lwjgl.opengl.GL11;
 
 import com.hbm.lib.RefStrings;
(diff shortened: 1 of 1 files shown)


HbmMods:
> i don't exactly like this solution, especially when it involves running a command.

I have reason to suspect that this will get the mod nuked off of curseforge instantly

Voxelstice:
would ``org.lwjgl.Sys.openURL`` still get it nuked
it literally does the exact same thing

BallOfEnergy1:
Make sure that both lwjgl2 and lwjgl3 have that function, or else you risk incompatibilities between NTM and GTNH's LWJGL3ify.

Voxelstice:
only LWJGL2 appears to have that function unfortunately
the better solution to this problem honestly would be a multi-line text editor in game instead of opening the OS' text editor but as far as im aware NTM doesn't really have one. book and quill's code could be inherited and modified

HbmMods:
the entire point is to just use the OS's editor because making a proper editor ingame sucks ass and literally any notepad program is better anyway

HbmMods:
honestly if opening the file fails it'd be sufficient to at least open the folder the file is in

Voxelstice:
Found a different solution that works with both LWJGL2 and LWJGL3 (only opens the folder)

PR #2938, opened 2026-06-04
consistent fluid texture pixel sizes for the new BAT window

legendarydoge30:
added a scaling factor for the fluid window textures on the new BAT
its scales up the pixels so theyre consistent with the way every other texture in the game looks

imgaes! :
image
image

BEFORE
image


churginator:
oh god he changed the tab format

legendarydoge30:
my stupid thing does it automatically idk if i can turn it off

PR #2944, opened 2026-06-05
Overlay

DTK-MrRaccoon:
added overlay for centrifuge crystallizer electrolyser

### DIFF
--- a/src/main/java/com/hbm/blocks/machine/MachineCentrifuge.java
+++ b/src/main/java/com/hbm/blocks/machine/MachineCentrifuge.java
@@ -5,15 +5,23 @@ import com.hbm.main.MainRegistry;
 import com.hbm.tileentity.TileEntityProxyCombo;
 import com.hbm.tileentity.machine.TileEntityMachineCentrifuge;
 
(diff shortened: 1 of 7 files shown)


HbmMods:
why, though

DTK-MrRaccoon:
big factory something is broken or stuck and now have to open 100 of machines to find it

legendarydoge30:
maybe this could be a client config thing
feels a little cluttered to be on by default but useful utility to whoever might need it
can be turned on and off on the fly

TheVitya2127:
Holo Inventory NTM Edition

DTK-MrRaccoon:
alright it has a config now default is off

HbmMods:
I'm more concerned about syncing this much extra data even when it's actually turned off clientside. Not a fan of this from a consistency standpoint either, these few machines have it but everything else doesn't. Plus, with centrifuges you can tell almost immediately if they run or not based on the sound, even in a dense noisy machine room you can tell easily by proximity. Assuming you haven't muffled absolutely everything, that is.

PR #2962, opened 2026-06-11
chicago pile thorium

DTK-MrRaccoon:
thorium support for chicago pile


HbmMods:
Don't believe that's necessary, the new chicago pile comes right after the pneumatics stuff I'm working on so the two block IDs we have to use for this will become obsolete almost immediately anyway.

PR #2971, opened 2026-06-19
DFC explosion radius bug

LasseTheBabo:
(don't mind the "fuck gradle" commit, this is my first pr)

### only change is in TileEntityCore.java (line 86-89)
On powerful double-core DFC setups (power output > 1PHE/s) it can happen that the product of "fill * mod" gets over the upper integer limit (Integer.MAX_VALUE).
This results in the sign bit being flipped and the product getting negative.
The size clamp in line 89 then gives the minimum radius of 50 blocks even if it should be 1000.

Casting temporary the product of "fill * mod" into a long should fix this because so it shouldn't flip negative at high values.

### result of a custom dfc simulation before the change
Uhh i hope you understand the result of this crap piece of my simulation code. You can test the setup it in Minecraft.
- two cores in a row
- all fuel types set to Oxygen
- both cores have Eye of Harmony
- Emitter power set to 100%
- explosion size results in -5827 without the clamp. In Minecraft it turns into 50 blocks radius
grafik

because github wont tell me how tf i exclude commits, the (half) dupe fix is also here

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -121,6 +121,8 @@ dependencies {
 }
 
 processResources {
+	duplicatesStrategy = DuplicatesStrategy.INCLUDE
+
 	// this will ensure that this task is redone when the versions change.
 	inputs.property "version", project.version
 	inputs.property "mcversion", project.minecraft.version
@@ -133,8 +135,8 @@ processResources {
 		filesMatching('mcmod.info') {
 			// replace version, mcversion and credits
 			expand([
-					version: version_name,
-					credits: project.credits
+				version: version_name,
+				credits: project.credits
 			])
 		}
 	}
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
--- a/src/main/java/com/hbm/inventory/gui/GUIToolBox.java
+++ b/src/main/java/com/hbm/inventory/gui/GUIToolBox.java
@@ -35,8 +35,8 @@ public class GUIToolBox extends GuiContainer {
 
 		// if the open box has changed or disappeared, close the inventory
 		} else if(this.mc.thePlayer.getHeldItem() != firstHeld) {
-			//this.mc.thePlayer.closeScreen();
-			//return;
+			this.mc.thePlayer.closeScreen();
+			return;
 		}
 
 		super.drawScreen(x, y, interp);
--- a/src/main/java/com/hbm/tileentity/machine/TileEntityCore.java
+++ b/src/main/java/com/hbm/tileentity/machine/TileEntityCore.java
@@ -36,7 +36,7 @@ import net.minecraft.util.Vec3;
 import net.minecraft.world.World;
 
 public class TileEntityCore extends TileEntityMachineBase implements IGUIProvider, IInfoProviderEC {
-	
+
 	public int field;
 	public int heat;
 	public int color;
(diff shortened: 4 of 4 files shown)


irisfalke:
you touched the sacred build system only god may save you

build.gradle
```diff
@@ -121,6 +121,8 @@ dependencies {
 }
 
 processResources {
+	duplicatesStrategy = DuplicatesStrategy.INCLUDE
```
HbmMods, line 124: so like what does this do

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
```
HbmMods, line 6: absolutely the fuck not

src/main/java/com/hbm/inventory/gui/GUIToolBox.java
```diff
@@ -35,8 +35,8 @@ public void drawScreen(int x, int y, float interp) {
 
 		// if the open box has changed or disappeared, close the inventory
 		} else if(this.mc.thePlayer.getHeldItem() != firstHeld) {
-			//this.mc.thePlayer.closeScreen();
-			//return;
+			this.mc.thePlayer.closeScreen();
```
HbmMods, line 38: you didn't mention this part in the actual PR

LasseTheBabo: i dont know exactly how prs work so i committed and wanted to make a second pr but idk how the fuck it works. when i changed that line the dupe was half fixed

build.gradle
LasseTheBabo: idk how i can exclude commits from prs and i changed  that for me (thats why i wrote dont mind "fuck gradle" commit) because intellij idea gave errors

LasseTheBabo:
uhh anyways, is this bug understandable? i stumbled over it because of developing a dfc bomb but the result wasnt the expected one so i debugged

build.gradle
MartinTheDragon: basically this allows multiple files with the same paths to be "combined". what combined actually means differs depending on context, and given it's just the `processResources` task, will probably just end up in one duplicate file randomly winning in the task's build output (it can make sense for zip files since they do allow duplicate entries, but that comes with other issues, so this setting has been a bad bandaid fix that spread across gradle users that you shouldn't really be doing unless you know why).
frankly this task shouldn't even have duplicates in the first place i think, so if this is needed to work, something strange is probably happening that might be related to the newer gradle version.

src/main/java/com/hbm/tileentity/machine/TileEntityCore.java
```diff
@@ -60,31 +60,34 @@ public String getName() {
...
 				int fill = tanks[0].getFill() + tanks[1].getFill();
 				int max = tanks[0].getMaxFill() + tanks[1].getMaxFill();
 				int mod = heat * 10;
-				
-				int size = Math.max(Math.min(fill * mod / max, 1000), 50);
-				
+
+				// fill * mod can create a value higher than 2^31 (max integer value)
+				// where at that point radius would go negative
+				long radius = (long) fill * mod / max;
```
MartinTheDragon, line 88: would `mod * (fill / max)` or `fill / max * mod` not work? it achieves effectively the same thing because your calculation can technically still grow above the maximum int32 value and become negative when the long is narrow-casted back into an int32. not that it's likely to happen in this case but it could be simpler, is all.

src/main/java/com/hbm/inventory/gui/GUIToolBox.java
MartinTheDragon: you have to create a separate git branch and commit to that one instead. any commits to the same branch will continue ending up in the same pull request.
also what does "half fixed" even mean?

MartinTheDragon (commented):
saving bob a little time but this pull request is too high effort to actually use and cherry-pick for what needs to be a one-line change which could've just been part of a github issue, sorry

RedstoneParkour:
> idk how i can exclude commits from prs

you'll have to do some fancy git stuff
1. make a new local branch
2. cherry pick the commits you want from the old branch into the new branch using `git rebase -i <old> <new>`, see also https://git-scm.com/docs/git-rebase#_interactive_mode
3. force-push new onto the remote-old (`git push --force-with-lease <remote> <new>:<old>`), updating the pr to point to the correct commit graph

alternatively you can redo your changes in a new branch and open a new pr

src/main/java/com/hbm/tileentity/machine/TileEntityCore.java
LasseTheBabo: could work too. didnt test it yet

LasseTheBabo:
thanks for the github instructions and by "half fixed" i mean the dupe happens when the gui is opened but because of network delay the swap (moveRows()) _can be executed after opening gui_. the closescreen thingy should be executed when the toolbox inventory isnt the same anymore to kill the ghost items but somehow it doesnt do it always

PR #2974, opened 2026-06-20
platemetal, new stalac/stalags, flesh block sounds, 2 new structures, and uhh CT block enum class

irislgtm:


irisfalke:
DENIED!!

irislgtm:
I Ate you.

NoErnesto:
freaky

Lukec2009:
did you add a new line in "src/main/java/com/hbm/inventory/recipes/ChemicalPlantRecipes.java" on purpose?

irislgtm:
yea so there was a rusted variant for the platemetal block which had a recipe, i removed that, must have been left over on accident

irislgtm:
@Lukec2009 ^^^

HbmMods:
can i see what those structures actually look like

irislgtm:
Screenshot 2026-06-22 163745
Pump

image
Dish

src/main/java/com/hbm/blocks/BlockEnums.java
```diff
@@ -81,4 +86,21 @@ public static enum LightType {
 		HALOGEN
 	}
 
+	public static enum PlatemetalType {
```
HbmMods, line 89: why no brown

src/main/resources/assets/hbm/lang/en_US.lang
```diff
@@ -5644,6 +5659,7 @@ tile.lightstone.bricks_chiseled.name=Lightstone Chiseled Bricks
 tile.lightstone.chiseled.name=Chiseled Lightstone
 tile.lightstone_tile_stairs.name=Lightstone Tile Stairs
 tile.lightstone_bricks_stairs.name=Lightstone Brick Stairs
+tile.lox_barrel.0.name=LOX Barrel
```
HbmMods, line 5662: whar

src/main/java/com/hbm/blocks/BlockEnums.java
irislgtm: brown looks like shit and we might need a rusted variant in the future

src/main/resources/assets/hbm/lang/en_US.lang
irislgtm: i migrated the barrels to enum before but reverted it cause i thought i would do it properly later 
lang is from that i guess

HbmMods:
(the PR is good you literally just have to throw out the unused lang stuff you added)

PR #2992, opened 2026-07-05
Fixed Radio Localization

TheVitya2127:
Localization didn't work because of the extra `rec`


TheVitya2127:
I also send Greetings to everyone from the Official NTM Discord Server.

HbmMods:
The radio is internally known has `radiorec`, so it's the container that's wrong, not the localization.

TheVitya2127:
Yes, I noticed it at night, sorry.

PR #2995, opened 2026-07-07
Fat Mine Remodel

TheVitya2127:
The model is quite old, so I decided why not redo it.
Now the model looks like this:

изображение


Bufka2011:
This is peak

HbmMods:
grafik

PR #2996, opened 2026-07-07
cyclotron remodel

legendarydoge30:
cyclotron remodel
tell me if i should tweak anything about it

image


TheVitya2127:
No way. Its really **PEAK**

RandomDude9587:
thats pretty fire

JameH2:
she cycl on my o tron

RandomDude9587:
till i antimatter

ASTARYOU200:
Yo that's peak af

HbmMods:
I may tinker with the color a little bit because I don't think the fusion color scheme works on a pre-RBMK machine but otherwise it looks really fucking nice

legendarydoge30:
i thought the same for the texture, i experimented with some other palettes but i couldnt get it to look the way i wanted it to so i defaulted to the fusion colors
i wait to see what you cook :p

sten104:
would you be intrested in making a texture for my shredder PR?

HbmMods:
grafik

reduced saturation on the top cables because they were a bit too in-your-face for elements that don't actually do anything, replaced all the metal panels with unpainted grey with the exception of the outer rings which are yellow, maybe even plastic, who knows

wasn't too sure at first but the design kind of grew on me, might reuse the scheme for more pre-fusion stuff

TheVitya2127:
It remains to fix the ports and the cyclotron is ready

PR #3007, opened 2026-07-12
Barrel improvement

TheVitya2127:
Like some players, I don't like the texture of the Ports on the Barrels, so I decided to Fix it a bit and Fix the UV Ports at the same time.

изображение
изображение
<details>
<summary>Now it looks something like this:</summary>

изображение
изображение

I also updated the Texture of the Magnetic Antimatter Container. Now it's in the Fusion style. I decided not to change the recipe of the barrel and left it up to you.
</details>


HbmMods:
yum, bscco barrel

PR #3014, opened 2026-07-14
Let's Make the Canisters and Gas Tanks Useful

TheVitya2127:
For a long time, it was considered strange to me that Canisters and Gas Tanks were in NTM, but they were absolutely useless against Universal Fluid Tanks.

I decided to fix it and make that all these tanks are used.

## Changes
- Canisters and Gas Tanks are now being craft in 4 pieces instead of 2
  - But Universal Fluid Tanks began to be craft in 4 pieces instead of 8, which makes them more __Expensive__, but used for all types of Fluids
  - Hazardous Material Tank now craft 1 piece instead of 4 (this is due to the fact that the recipe uses 1 Universal Fluid Tank (they are extremely rarely used at the moment, seriously) )
- Canisters and Gas Tanks can now be Recycling
- Universal Fluid Duct are now made in 12 pieces, and BoxDuct in 2 pieces
  - This change is due to making them both user-friendly and (almost) the same price
  - Because... um.... The Cables BoxDuct and Exhaust BoxDuct are much more craftable than Regular BoxDucts
  - Universal Fluid Ducts can now also be Recycled
## Fixed
- Fixed a Bug that prevented Old Big Ass Tank 9000 from being Recycled
  - I do not know why I Fixed it, Lmao

I feel like I'm going to be crucified for such changes :smilecat:

### DIFF
--- a/src/main/java/com/hbm/inventory/recipes/anvil/AnvilRecipes.java
+++ b/src/main/java/com/hbm/inventory/recipes/anvil/AnvilRecipes.java
@@ -560,9 +560,9 @@ public class AnvilRecipes extends SerializableRecipe {
 		constructionRecipes.add(new AnvilConstructionRecipe(new ComparableStack(ModItems.billet_pu238be, 1), new AnvilOutput(new ItemStack(ModItems.plate_fuel_pu238be))).setTier(4));
 
 		for(int i = 0; i < 15; i += 3) {
-			constructionRecipes.add(new AnvilConstructionRecipe(new OreDictStack(IRON.plate(), 1), new AnvilOutput(new ItemStack(ModBlocks.fluid_duct_box, 1, i))).setTier(2).setOverlay(OverlayType.CONSTRUCTION));
-			constructionRecipes.add(new AnvilConstructionRecipe(new OreDictStack(CU.plate(), 1), new AnvilOutput(new ItemStack(ModBlocks.fluid_duct_box, 1, i + 1))).setTier(2).setOverlay(OverlayType.CONSTRUCTION));
-			constructionRecipes.add(new AnvilConstructionRecipe(new OreDictStack(AL.plate(), 1), new AnvilOutput(new ItemStack(ModBlocks.fluid_duct_box, 1, i + 2))).setTier(2).setOverlay(OverlayType.CONSTRUCTION));
+			constructionRecipes.add(new AnvilConstructionRecipe(new OreDictStack(IRON.plate(), 1), new AnvilOutput(new ItemStack(ModBlocks.fluid_duct_box, 2, i))).setTier(2).setOverlay(OverlayType.CONSTRUCTION));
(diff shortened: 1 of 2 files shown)


nuhyuh:
another day another banger at the pull request tab

TheVitya2127:
Sorry. I'm trying to explain in as much detail as possible why I made certain changes.

nuhyuh:
> Sorry. I'm trying to explain in as much detail as possible why I made certain changes.

i mean thats all fine im just saying another amazing PR

NoErnesto:
wouldn't just getting rid of universal tanks in 528 be a better idea?

NoErnesto:
This changes you make honestly suck, even if universal tanks are more expensive, its not gonna do anything just delay getting a stack in your inventory and never crafting them again. If you want to "rebalance" the fluid containers just get rid of universal tanks and call it a 528 feature. Or lock the universal tanks behind something, like saturnite, but as i said that just delays getting a stack of them and never crafting them again.

TheVitya2127:
> This changes you make honestly suck, even if universal tanks are more expensive, its not gonna do anything just delay getting a stack in your inventory and never crafting them again. If you want to "rebalance" the fluid containers just get rid of universal tanks and call it a 528 feature

I can't understand what the problem is and why it needs to be moved to 528 mode.

> Or lock the universal tanks behind something, like saturnite, but as i said that just delays getting a stack of them and never crafting them again

I can't just add Saturnite to the recipe even in 528 mode, because not all types of fluids can be poured into canisters and gas tanks

For the most part, this is simple dissatisfaction due to balance changes, which players will sooner or later get used to. You will be crafting universal fluid tanks anyway

TheVitya2127:
You can get rid of canisters or gas tanks at any time by simply recycling them (anvil)

NoErnesto:
The problem is, people are going to craft the universal tank and ignore all other means of fluid containers. Simply making it more expensive doesn't do much, if the ease of use is on the table. As to 528 mode, I'm not talking about moving the feature, but totally getting rid of the functionality of universal tanks. For fluids that dont go in anything else, you could keep the universal tank item, and restrict all the other fluids to their respective containers. Saturnite was just a loose suggestion, it could be anything thats gated behind something if you want to go this way, which I also said is not a good idea. And finally, yes, I will be crafting the universal tanks anyway, and thats the issue, because this change doesn't actually change anything.

NoErnesto:
I have no issue with a thing like that not even being a 528 feature, its just the fact that universal tanks can hold nearly anything that makes them infinitely superior over alteranives no matter how expensive it is (making any cost changes, well, meaningless)

TheVitya2127:
> I have no issue with a thing like that not even being a 528 feature, its just the fact that universal tanks can hold nearly anything that makes them infinitely superior over alteranives no matter how expensive it is (making any cost changes, well, meaningless)


Then, in principle, it is possible to change their number from 4 to 1 :trollface:

I did what I could. I just suggested it as an idea. This PR will probably be closed, so I don't care.

NoErnesto:
Think like, poison capsules from factorio. Sure you could use them, there are people using them, but they are mostly useless compared to like a tank. More expensive, but people use it anyway.

NoErnesto:
Yeah, 4 to 1 would be funny. I'd still prefer those over any other container tho.

HbmMods:
nah

PR #3017, opened 2026-07-15
Add /ntmwikirender command to render screenshots of items and blocks

### LINKED ISSUE #3013: Make the `GUIScreenWikiRender` functionality usable in a command
the functionality of this file should be entirely usable within a command, so you can use it even with a built jar.
it would be like so:
`/ntmwikirender <internal type of items to render>`

iamnotmega:
the command is used like so:
`/ntmwikirender <type>` where `<type>` is the internal type of the item(s) you want to render, such as `MachineIndustrialTurbine` or `ItemGunBaseNT`.
this behaviour means that it will be intended **SOLELY** for wiki editors and developers
the command, however, doesn't support rendering specific ITEMS, only the `Item` type itself

closes #3013

### DIFF
--- a/src/main/java/com/hbm/commands/CommandWikiRender.java
+++ b/src/main/java/com/hbm/commands/CommandWikiRender.java
@@ -21,6 +21,7 @@ import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.ItemStack;
 import net.minecraft.command.ICommandSender;
 import net.minecraft.util.EnumChatFormatting;
+import net.minecraftforge.client.ClientCommandHandler;
 import net.minecraft.client.Minecraft;
 import net.minecraft.command.CommandBase;
 import net.minecraft.item.Item;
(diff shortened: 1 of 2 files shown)


MellowArpeggiation:
neato, but can we get an option to specify slot scale? would be nice to do 32px slot renders and big 512 machine renders with the same command

iamnotmega:
alright, ill try to see what i can do
`/ntmwikirender <type> <slot scale>` perhaps?

MellowArpeggiation:
makes sense to me, tab completion for the type would also be useful

HbmMods:
that's a serverside command that invokes `Minecraft.getMinecraft()`, we are so not doing that

iamnotmega:
if this implementation gets merged and is stable enough and preferred then i personally think the `GUIScreenWikiRender` class and the `LCTRL + 0` keybind should be deprecated or even removed entirely
i dont know just a thought

iamnotmega:
> neato, but can we get an option to specify slot scale? would be nice to do 32px slot renders and big 512 machine renders with the same command

im unsure if thats even possible so for now ill just not implement it

HbmMods:
is `register()` even called anywhere?

iamnotmega:
> is `register()` even called anywhere?

i dont think it is, probably my fault cause i didnt make it client side only properly

iamnotmega:
yeah after that failed attempt turns out the command doesnt even get LISTED anymore
i think i should probably just move it back into the mainregistry or what should i do

iamnotmega:
i moved it back

HbmMods:
grafik

PR #3020, opened 2026-07-16
i hope this commit wont destroy whole ntm

Alexmaster75:
smooth gauges everyone, finally in texture and linear taste!

heres how they look in-game:

should be noted that the linear gauge can be configured to be at ANY angle, maintaining the scale nonetheless. this allows for vertical or even diagonal gauges.
the texture one instead is basically only made for the ccgt, but it has the base to be potentially full 360°.
the main reason for this commit is to get rid of the need for external texture gauges (i still have nightmares of ccgt's....)


BabyPunchingMarathon:
amazink

RandomDude9587:
I hate these

HbmMods:
y u m

PR #3026, opened 2026-07-18
Added a lot of RoR support to the battery socket and PA, added extra ports to PA, updated calculator

legendarydoge30:
ough part 2

yeah i added that thing there "max power ror reader option for battery sockets"
called it "maxfill"
i thought maybe "capacity" but thats like capacitor stuff and batteries also exist so maxfill is better i think

i tested batteries and capacitors and items and they work it seems yay


legendarydoge30:
image

legendarydoge30:
i can now make the battery monitoring program fully automatic after its been set up and the batteries can be switched out and updated without making any changes to the code!!!!

legendarydoge30:
ok so i just added lots of RoR support to the Particle Accelerator and i didnt realize i was on the same branch so i guess that merged my PRs, ill update the title

legendarydoge30:
i still have more things to change to the PA pls dont merge

legendarydoge30:
ok tested everything, seems to work
image


added all of those readable values to the particle source and for every other component i added "temperature" "pfm" and "pfmcold"
i wouldnt have included pfm and pfmcold since temperature is probably the only thing you need but someone said their OC program relies on reading the PFM buffers so i just kept it it becuase why not

legendarydoge30:
+added modulo to calculator (x%y)

legendarydoge30:
i was also going to add binary functions like min, max, and whatnot but then i looked at that token thing and decided not to even try

legendarydoge30:
found out that the particle source does not have enough ports for everything! its 7 out of 8 ports, so ill add an extra concealed port on the bottom and update the texture

legendarydoge30:
people want both OC *and* RoR.. . ,  so ill add 3 ports., , at the bottom  ..  ,,

legendarydoge30:
i put more ports on the PA now
image

it took way too long, they all seem to work

legendarydoge30:
thats enough ports for all the cooling, power, both ror controller and reader, particles, capsules, AND OC, and you have a spare port

legendarydoge30:
hold on its broken

legendarydoge30:
fixed

PR #3033, opened 2026-07-21
New Pylon Model And SpinSocket

TheVitya2127:
The New VityaSlop™

images

## Changed
- New Industrial Electricity Pylon Model
  - The recipe has also been changed
изображение

изображение

- Any energy items are now displayed in Battery Socket
изображение

Despite the fact that I have conducted tests and everything is working fine, I **strongly** recommend checking out the code, as I am very worried that I have implemented it crookedly.

And a tiny fix from the previous pull request (#3022)


NoErnesto:
is it now a multiblock? or is it still like it was (only the bottom is an actual block)

nuhyuh:
nuclear rework mod

TheVitya2127:
> is it now a multiblock? or is it still like it was (only the bottom is an actual block)

Yup

Без имени

InstantNootles:
Forcing people to use steel to craft basic power pylons in the early game when wooden medium power poles exist alongside a steel counterpart is a very strange choice. That and the weird ribbed core on a sloping pyramid strut layout along with the trusses that end abruptly just looks kinda bad to be honest.

TheVitya2127:
> Forcing people to use steel to craft basic power pylons in the early game when wooden medium power poles exist alongside a steel counterpart is a very strange choice. That and the weird ribbed core on a sloping pyramid strut layout along with the trusses that end abruptly just looks kinda bad to be honest.

a-train-a-train-the-boys

So the current version is better than what I suggested, right? The design of the pylon itself was taken from the factorio. I can just take it, repaint it to look like wood, and change the recipe if you want.

NoErnesto:
or it could be a slightly better one, like 35m or something

TheVitya2127:
изображение

whu

BabyPunchingMarathon:
tbh i think it looks better than the old one, both the steel and wooden models. only thing i'd change is the craft since the old one was dirt cheap and gave you four of them pylons, either the craft should be less expensive or all the other plyon recipes rebalanced otherwise its more worthwhile to just use the other pylons, but thats just my 2 cents

nuhyuh:
honestly imo
do both the steel and the wood one (but please just keep the truss beam crosses for the entire thing)

TheVitya2127:
Okay, when I get home, I'll just **Destroy** this wretched Pylon. It's starting to get on my nerves with its imperfections.

legendarydoge30:
Model looks good but I don't like the fact that it is a multiblock now
Though I think making those diagonal beams a full pixel thick is a strange choice. It would be more efficient to make them thin planes and you also wouldn't need to stagger them like they are right now. (Or merge them into one full piece if you want to deal with that topology)

nuhyuh:
> Model looks good but I don't like the fact that it is a multiblock now

i mean whats the bad thing in that

legendarydoge30:
You could make custom pylons or even hide pylons within blocks when they're not a multiblock. Making it a multiblock just removes one degree of freedom for building that used to exist in previous versions.

TheVitya2127:
Just use a Super Connector, Lmao

legendarydoge30:
Post-Technetium and much more expensive... that's not a good solution. It also does not look that good if you're not using it for big pylons, the connector is too long.

TheVitya2127:
What the hell difference does it make what it looks like if you're going to bury it in blocks anyway?

legendarydoge30:
I literally said the model looks good? I am not against the remodel in any way, I only pointed out the fact that making it a multiblock is just a downgrade from what it is right now. Remove the extra blocks, keep it a 1x1, and it's fine.

TheVitya2127:
_Why do people so much like old things
For many years, players often told me about the possibility of adding a new model and adding a multiblock, but when I did this, everyone strongly disliked it.

I don't know what to do or how to solve these problems. Let me get home and I'll delete these fucking changes.

legendarydoge30:
I guess reading comprehension is not something they teach in school anymore so let me put in pictures for you.

pylons

TheVitya2127:
Bruh

TheVitya2127:
To avoid waiting for Bob's answer, I'll answer for him:
 Nah
I'll create a new PR a little later. Maybe in a week.

legendarydoge30:
G3vrrbjXEAA7wGm

TheVitya2127:
I really want to sleep.

PR #3040, opened 2026-07-23
Machine Placement Overlay 

WolfEclipses:
Yeah we talked about this before you disappeared, its done now.
Key Notes:

-It is all in the dummyblock file
-No noticeable performance drops (Im stuck on 20fps anyway)

-All machines that render multiple boxes, mainly XLMachines need a separate function that declares all the bounds they have.
-I put those in for all current XL machines, and checked that every machine in the machine tab works and is as accurate as necessary.

-Mellow gifted part of the rendering code, thanks mellow <3

-Currently there is no keybind to toggle this off, I was not sure how you wanted that handled and well, you are gone so...
The ideas I thought of: 
    -Keep as is, always on
    -Only activates when crouching, or holding some other keybind
    -Is under the HUD toggle, so it activates when HUD is on, and turns off when HUD is off
    
Meow Meow Meow Meow Meow Meow


solosimz123-blip:
meow mrrrr meoa meow mrrrr

HbmMods:
grafik

Issue #2383, opened 2025-08-23
[Render Bug] All laser effects causes weird visual issues with multi blocks and custom model objects (e.g. missile)


NLWaterL:
All laser effects include FEL, ICF laser and Laser rifle's laser.

Just place a Multiblock / Missile behind a laser beam, and then the laser beam would either let you see through that multiblock, or the multiblock's model would "overlap" the laser, hiding the laser "behind".

FEL causes a see-through while ICF and laser rifle causes a render layering issue.

I didn't install any optimization mod.

Also, why the force field cannot block laser rifle?

HbmMods:
That's an issue in vanilla due to rendering order (or rather, lack thereof), you'll see effects like that in a lot of places where tile entities use transparency. Not a whole lot I can do about that.

PR #3044, opened 2026-07-24
Fix beam render order against custom models and multiblocks

NLWaterL:
This PR fixes the issue #2383

Beams rendered during world rendering are now queued and replayed during RenderWorldLast with depth writes disabled, after the depth buffer has been fully updated. This fixes the rendering issues regarding beam effects. FEL lasers no longer create see-through holes in multiblocks, and other beams effects no longer disappear behind machines that are actually behind them (ICF, laser guns...).

The BeamPronter use cases for the world-in-a-jar remain unaffected. These changes have been **tested in-game and work as expected**.

**One known tradeoff** is that particles now always render behind beams, regardless of their actual 3D positions. However, this is an acceptable compromise, as the original model rendering issues are both more noticeable and more common than this particle ordering problem.

### DIFF
--- a/src/main/java/com/hbm/main/ModEventHandlerClient.java
+++ b/src/main/java/com/hbm/main/ModEventHandlerClient.java
@@ -44,6 +44,7 @@ import com.hbm.render.anim.HbmAnimations;
 import com.hbm.render.anim.HbmAnimations.Animation;
 import com.hbm.render.block.ct.CTStitchReceiver;
 import com.hbm.render.item.weapon.sedna.ItemRenderWeaponBase;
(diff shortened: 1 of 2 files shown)


NLWaterL:
Tested with optifine, also works

HbmMods:
grafik

SampoPro:
image

sludge

HbmMods:
god i fucking hate git rebase

@NLWaterL clanker code doesn't actually have an author, so it's not compatible with our license. shoulda read the contributing guidelines first. never do that shit ever again

PR #3046, opened 2026-07-25
Aeropuerto or something (PA Source GUI)

WolfEclipses:
-Meow

Here is a picture of the new one, functionality was obviously maintained.

image

This was suggested as an alternative to the current markers. It was also suggested for the detector, I would like your opinion on this.

image


HbmMods:
I like the colored ones

PR #3054, opened 2026-07-26
mercury battery

DTK-MrRaccoon:
mercury battery


HbmMods:
literally why

DTK-MrRaccoon:
mercury is kind of useless and does pile up so this could have been use for it

PR #3063, opened 2026-07-27
add fluiditemstorage classs

irislgtm:
this is for something I was planning to do in the future
this class just extracts the nbt access stuff from the pipette class, and re-implements them in a separate class so it can be used easier by other items (one which i was planning!) and makes the base pipette use it.

It will be useful for future items i think.

### DIFF
--- a/src/main/java/com/hbm/inventory/fluid/tank/FluidItemStorage.java
+++ b/src/main/java/com/hbm/inventory/fluid/tank/FluidItemStorage.java
@@ -1,0 +1,30 @@
+package com.hbm.inventory.fluid.tank;
+
+import com.hbm.inventory.fluid.FluidType;
+import com.hbm.inventory.fluid.Fluids;
+import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTTagCompound;
+
+// This is a class for items that use pipette like behaviors (rn its used for the pipette and autodrip)
+public class FluidItemStorage {
+	public static void initNBT(ItemStack stack) {
+		stack.stackTagCompound = new NBTTagCompound();
+		stack.stackTagCompound.setShort("type", (short) Fluids.NONE.getID());
+		stack.stackTagCompound.setShort("fill", (short) 0);
+	}
+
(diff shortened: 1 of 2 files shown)


HbmMods:
Wouldn't it be easier to either add the static helper methods into `IFillableItem` or as default implementations rather than a completely separate util class? Also there's a few other partially fillable items like all the fueled tools and armor pieces, the chemthrower and the welding torches.

irislgtm:
Yeah that is true but i didnt want to convert all the fillable items just yet, the class was just for the two specific behaviors that Pipette and my other item have.. If you want i could try to migrate the other items to this system, then i think i'll add them into `IFillableItem` as you said

irislgtm:
Also the "other item" i was talking about was this 
Screenshot 2026-07-27 153557
Screenshot 2026-07-27 153440

irislgtm:
I was gonna replace the [Delicious] FT with a more generalised [Consumable] FT that applies regen based on the food value in the Mist / Chemthrower, and can be used in the Auto Drip (which is like a minefactory reloaded autofeeder, it just drips fluids into your bloodstream, with different effects) the class was so i can add the filling behavior for that easier rather than just repeating what pipette alr does

irislgtm:
the sludge..

irislgtm:
> Yeah that is true but i didnt want to convert all the fillable items just yet, the class was just for the two specific behaviors that Pipette and my other item have.. If you want i could try to migrate the other items to this system, then i think i'll add them into `IFillableItem` as you said

So, should i do that or what?

HbmMods:
ye

irislgtm:
im removing this anyway ill do a pr later

irislgtm:
@HbmMods BTW, after a bit i've seen that the other partially fillable items like the welding torches and stuff wouldn't really benefit from the new stuff.. so is it just OK if i make a PR with my Auto Drip feature and the changes to `IFillableItem`? It would also entail removing the [Delicious] FT as i said before, I was just asking again so i could be clear that all of this is OK

PR #3078, opened 2026-07-29
Me when there are more h's then there should be (Cracked Naphtha Qmaw)

WolfEclipses:
Insert funny description here


HbmMods:
I've noticed this in some prior PRs, you're mentioning oil 3 stuff *before* the actual important bits. also "It can also be fractioned." sounds like that's a side thing even though it's one of the main uses. The page should ideally also mention diesel directly.

WolfEclipses:
It has been fixed, thanks for your feedback on that, I didn't even realize I was doing it and now I cant unsee it, so I will be fixing it in the future, try and see if I did it already in a particular fluid

PR #3098, opened 2026-08-02
Anvil Redesign pt 1

WolfEclipses:
-Not a rework
-Adds a button to sort by different categories so we can find certain things faster
      -Ex: We want to recycle a bunch of loot and can now do that faster
-Search still searches ALL items not just the type selected, clear search box to go back to sorting    
  
-Reason it is pt 1 is Im going to update gui textures next, did not want to combine both in case you hate this

-Ill gladly move location of button or order or anything just let me know

-Meow


HbmMods:
Instead of using a magic number, couldn't this be an enum?

WolfEclipses:
This has been updated to Enums, I lowkey forget enums exist the second im not using them and defaulted to my state machine experience with verilog T_T

PR #3099, opened 2026-08-02
Multiblock Shape Refactor

WolfEclipses:
DO NOT MERGE THIS. IT WILL BREAK EVERYTHING.

Background
     -The gui placement thing was hell because of all the different systems in place, when making that mellow and I talked about it and determined that it should be refactored/redesigned/simplified, this is the byproduct of that, it also is supposed to help with some space stuff??? (idk not my department)
     
What this does
     -Takes and unifies the dimension usage (checkPlacement, fillSpace, getAllDimensions, getDimensions) for all multiblocks into BlockDummyable, and allows us to define a shape of a multiblock in a more simple manner, it also eliminates a lot of overrides in every machine (it'll cut ~15-50 lines of repeat information from most multiblocks)
     -Some changes will be made once all are converted (Fixing getDimensions) (Limiting call access)
         -These werent done since converting only one was causing issues
     
BEFORE I go and rewrite every single multiblock in the game to this new system, (as that will take a shit ton of time) I wanted to show you so you could:
-Look at it
-Consider it
-Get acquainted with it
-Give input
-Take it for a walk (idk)

Currently, this ONLY works for the leviathan turbine as that has the most hellish multiblock of all time (Im looking at you front port)

This only really affects you and maybe a contributor once in a while, so please prioritize your happiness on this.

While I have tested this, and fixed the bugs I did encounter, there may be something I have missed, and I will continue testing as I go.

Thank you :P


TheVitya2127:
ReFactorio

PR #3107, opened 2026-08-03
If its so good, why isn't there High-Performance Solvent 2? (HPS QMAW)

WolfEclipses:
yo sometimes I just want to sleep and then make a qmaw
somtimes is every day...
Hi daily PR readers :D


MaybeOneMoth:
Hello daily PR releaser

HbmMods:
`[[breeding|Fusion Reactor Breeding Chamber]]` is not entirely correct since it's primarily an irradiation column recipe that's also doable later in the fusion reactor. Should also probably mention that in both recipes it's post-RBMK.

WolfEclipses:
Fixed, changed  to rbmk irradiation chamber. (Btw the qmaw for that doesnt exist) and clarified rbmk is required to make this through either method

HbmMods:
`to produce due to the prerequisite of [[bismuth|Bismuth]] for [[reformate gas|Reformate Gas]]` is also not entirely correct since the chlorinated petroleum wax way doesn't actually need refgas at all.

WolfEclipses:
Clarified the uses are not necessarily correlated (from your last comment) also switched it to list oil3 as the template now that that exists, I think that's the right reference unless one for bedrock ore appears

PR #3108, opened 2026-08-03
add ror support for cargo elevator

APie357:
adds redstone over radio support to the cargo elevator, and lets you manually switch the direction of the elevator while in motion

for the people who update their worlds: this might send elevators that are already extended to the bottom when updated, but idk

how 2:
send `setextension!<height>` via ror to an ror controller for the elevator

### DIFF
--- a/src/main/java/com/hbm/tileentity/machine/TileEntityCargoElevator.java
+++ b/src/main/java/com/hbm/tileentity/machine/TileEntityCargoElevator.java
@@ -2,6 +2,7 @@ package com.hbm.tileentity.machine;
 
 import java.util.List;
 
+import api.hbm.redstoneoverradio.IRORInteractive;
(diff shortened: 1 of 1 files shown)


APie357:
https://github.com/user-attachments/assets/dc71cd80-338d-42b0-96c6-f2d8783f965b

FluffsADev:
This is splendid I will use this
image

HbmMods:
This is basically exactly how I envisioned the elevator would work. Instead of having dedicated floor segments or doors or something like that, just basic RoR that lets you do all of that instead.

theNTMenjoyer:
it said in this commit that cargo elevator has ROR support
it does
but i checked
and you can't place any torch anywhere to interact with the elevator
at all
not even with setblock
please fix



or not maybe i just don't know what to do to get the controller torch to work

PR #3110, opened 2026-08-03
Add a getplayername command to the Radar Satellite

FluffsADev:
Adds a command to the Radar Satellite that allows for a player name to be output if a radar entry is a player.

This is useful for displaying names of players found by a radar satellite and provides a non-changing string for the autocal to use for comparison unlike target ids which will change.

image


HbmMods:
Not intended for the radar satellite. The spy satellite can already do this, although limited to detecting all players within its view and only players on the surface.

PR #3111, opened 2026-08-04
They call this Ichigo the way its Bleach (Chlorine QMAW)

WolfEclipses:
Does it go in the pool or nah...

-I just realized how many fixes Im gonna have to PR once these get merged, there are likely referral issues, and I have many files I referenced that just dont exist so gonna need to make some machines/guide pages T_T


HbmMods:
The combination oven bit should probably mention glowstone, as it's like the number 1 earlygame way of making chlorine for cheap.

WolfEclipses:
Its fixed, I also removed oil1 and oil2  references since it has no relation to those, no use case or production method is from those or related closely.

PR #3115, opened 2026-08-05
Floor vs Antimatter (qmaw), who wins?

WolfEclipses:
Yeah, did not include dangerous drop as it literally is in the tooltip


HbmMods:
RBMK should be capitalized and linked

WolfEclipses:
Fixed, meow

PR #3116, opened 2026-08-06
Jetpack joke here (kerosene QMAW)

WolfEclipses:
ok so, this one I didnt cover that it could be used for high tier fuels or that it could be turned into petgas as those are in my opinion lesser uses, but if you want them in let me know.


HbmMods:
Fractioning comes first, so you should mention it first.

WolfEclipses:
Fixed, I have not done reformate gas or anvil requested changes yet, Im too tired

PR #3120, opened 2026-08-07
Mineral Clumps

polyvus:
Adds Mineral Clump

Mineral Clumps are obtained using the rock mill.
Crushing cobble stone has a 2% chance of giving one mineral clump,
which you can put 2 of them in a chemplant to get one of these items

- 30% chance gravel
- 15% chance sulfur
- 15% chance niter
- 10% chance iron powder
- 10% chance copper powder
- 10% chance lead powder
- 5% chance gold* powder
- 5% chance tungsten podwer
 (+100mB ore slop)

you can also just shred it to get gravel


sten104:
this would only be good in a skyblock and modpacks can already do this, i see absolutely no point in this

polyvus:
> this would only be good in a skyblock and modpacks can already do this, i see absolutely no point in this

Its just a cheap "B.O.P"
I just thought that if you have more power than you need and your still early game you can use this.
And I'm just thowing stuff at a wall and see what sticks, I don't expect anything..

TheVitya2127:
> Its just a cheap "B.O.P" I just thought that if you have more power than you need and your still early game you can use this. And I'm just thowing stuff at a wall and see what sticks, I don't expect anything..

This makes absolutely no sense and will break the balance.

The current BoP is already accessible early in the game. The recipes are incredibly cheap.
IMG_20260808_005047
IMG_20260808_005103


The new cobblestone refining recipe, which can be crafted almost out of thin air, is too unbalanced due to the huge yield chance (tested with basic sand and gravel refining in half an hour of play).

Additionally, Ore Slipe can be used for infinite BoP, where you don't have to dig new chunks, and also get infinite sulfur, niter and mercury without mining BoP.

polyvus:
> > Its just a cheap "B.O.P" I just thought that if you have more power than you need and your still early game you can use this. And I'm just thowing stuff at a wall and see what sticks, I don't expect anything..
> 
> This makes absolutely no sense and will break the balance.
> 
> The current BoP is already accessible early in the game. The recipes are incredibly cheap. IMG_20260808_005047 IMG_20260808_005103
> 
> The new cobblestone refining recipe, which can be crafted almost out of thin air, is too unbalanced due to the huge yield chance (tested with basic sand and gravel refining in half an hour of play).
> 
> Additionally, Ore Slipe can be used for infinite BoP, where you don't have to dig new chunks, and also get infinite sulfur, niter and mercury without mining BoP.

The recipes are insanely cheap but processing isn't.
I could make it so it doesn't give you ore slop and lower the chance / higher input energy for the chemplant
but other than that I don't see a problem.
And you call this unbalanced when you can make one astroid mining ship and it will give you infinite powders/crystals without needing to feed it anything (unless it got changed which I'm not aware of)

TheVitya2127:
> And you call this unbalanced when you can make one astroid mining ship and it will give you infinite powders/crystals without needing to feed it anything (unless it got changed which I'm not aware of)

Indeed. A Cobblestone made of air and an Asteroid Mining Ship, which is available after T3 oil, wow.

What is the problem with making the cheapest processing of bedrock ore? If you increase the power consumption, it will not make sense compared to BoP.

If were to add something similar to your idea, then could probably make an analog of "Ore Nitra", which I've been thinking about for a long time for NTM Space, but in a regular NTM there is no particular need for this (unless you are playing with an expensive mode)

TheVitya2127:
The problem is that Cobblestone is a very accessible material, there is always a lot of it and it can be produced infinitely even at the early stage of the game.

src/main/java/com/hbm/inventory/recipes/ShredderRecipes.java
```diff
@@ -219,7 +219,7 @@ public void registerDefaults() {
 		for(ItemStack sapling : saplings) ShredderRecipes.setRecipe(sapling, new ItemStack(Items.stick, 1));
 		for(ItemStack dust : lapis) ShredderRecipes.setRecipe(dust, new ItemStack(ModItems.powder_cobalt_tiny, 1));
 		
-		for(EnumBedrockOre ore : EnumBedrockOre.values()) {
+		for(EnumOre ore : EnumOre.values()) {
```
HbmMods, line 222: what are you doing here

HbmMods:
I'm not opposed to the idea, but the way it's implemented right now kind of conflicts with the idea behind the rock mill. Primarily, the rock mill is supposed to be a way to either automate or bulk process certain materials as a stepping stone for building a BOP setup, providing some extra materials that are either annoying to get or otherwise rare from regular mining without increasing ore yield or just making the ores way more common. Iron and copper were excluded deliberately because after you get a crucible, you basically never run out of those until you're way into midgame due to the giant hematite and malachite deposits.

> this would only be good in a skyblock and modpacks can already do this, i see absolutely no point in this

The secondary idea behind the rock mill was to provide an early and cheap machine that could replace ex nihilo. Indeed, using the config you can already assign any material you want to the recipes.

> This makes absolutely no sense and will break the balance.

Given that the outputs are mainly just common ores and need at the very least a chemical plant, I don't think balancing is the main concern here.

Issue #3114, opened 2026-08-05
I18nUtil Cascade


TheVitya2127:
To be honest, I haven't been able to reproduce the problem, but I'll post it anyway.

<img width="480" height="95" alt="Image" src="https://github.com/user-attachments/assets/b1b6735b-b2f5-483c-9c2e-058773a5c302" />

The error is probably caused on the server side because of these lines:

https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/9a2eb731adb90863063ba24ca551c890226fec09/src/main/java/com/hbm/items/tool/ItemPollutionDetector.java#L38-L40

It's not particularly difficult to fix it, *maybe*.
I'm fucking apologizing again AND DON'T YELL AT ME, I'M STILL READING YOURS MESSAGES FROM THE FUCKING SERVER.

BallOfEnergy1:
Did you test it on a dedicated server/client pair? That's the main way to reproduce these kinds of issues with I18n (client-side calls on server-side).

TheVitya2127:
> Did you test it on a dedicated server/client pair? That's the main way to reproduce these kinds of issues with I18n (client-side calls on server-side).

I have no idea how to do this. There is no documentation, so I do not know how to run the server-side tests.

BallOfEnergy1:
You start a dedicated server instance (using either the built-in IDE tools, you can look this up or use the gradle task `runServer`), or by plopping the built jar into a dedicated server instance. You can then join that server using a local client instance (with either `runClient` or otherwise). [This is part of the contribution guidelines, and you should have tested it on both client and server.](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/master/CONTRIBUTING.md#test-your-code)

TheVitya2127:
> [This is part of the contribution guidelines, and you should have tested it on both client and server.](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/blob/master/CONTRIBUTING.md#test-your-code)

Yes, I have read this file carefully. I use Sublime Text for all my PRs, so there may be problems.



> You start a dedicated server instance (using either the built-in IDE tools, you can look this up or use the gradle task `runServer`), or by plopping the built jar into a dedicated server instance. You can then join that server using a local client instance (with either `runClient` or otherwise).

Thanks, next time I will run tests from the server-side.

TheVitya2127:
[...]

PR #3121, opened 2026-08-07
Fix I18N server side call

Deeplerg:
Fix pollution detector detecting I18N server side calls

image


Deeplerg:
Closes #3114

TheVitya2127:
Thank you for releasing the PR with the correction of my crap faster than me.

PR #3125, opened 2026-08-09
Reformate gas qmaw

WolfEclipses:
Im tired


HbmMods:
I think instead of listing all the machines it can potentially come out of, it's more important to say that it's an end product from oil 3.

WolfEclipses:
Fixed

PR #3132, opened 2026-08-11
powered drill durability mechanics

tam100:
initially i thought this would be a +3 characters fix, but it turns out the drill does not have any behavior that depends on durability. i decided to add some myself, because it looked more like a mistake than intentional design, since the durability setter was never actually changing the durability.

the following things have been added:
- drill now decrements durability properly, 3x as fast when drilling mobs
- all drillbits have a new "wear" multiplier, which makes the drill wear down slower the higher the tier of the drillbit
- damage is now linearly decreased as wear increases, down to 75% damage (didn't think this one through, might need a rebalance)
- mining area shrinks into a noisy sphere starting from 75% durability and below, becomes impractical for mining at high enough wear, encouraging repair kits or better drillbits
- drill now shows a durability bar on the bottom right, just like any other gun

if none of this sits right with you, then at least please fix the wear increase in XFactoryDrill and turn off calcWear.


70000hp:
they were never meant to actually use durability because tool durability as a mechanic is pointless and the drills replace the terra drills

70000hp:
the PR seems to be an outright downgrade in functionality of the drills, for no particular gameplay purpose other than "realism" which ntm does not bother with to that degree anyway

TheVitya2127:
Realismio

Randomdude0689:
This is garbage

snarkysnappy1:
fau armor also has unused durability

TheVitya2127:
I love it when simple things are compared to late-game items.

Randomdude0689:
tam 69iq

TheVitya2127:
image-14

tam100:
> the PR seems to be an outright downgrade in functionality of the drills, for no particular gameplay purpose other than "realism" which ntm does not bother with to that degree anyway

if youre right then thats fine, i also like drills having infinite durability, it just looked more like an oversight the way durability was disabled in the code, so i thought itd be fun to implement.

i tested this myself for the record, it takes at least least 10 buckets of fuel for the mining effects to *start* to kick in, and theyre honestly tolerable until around 50 to 20% durability (depends on aoe)

i was honestly expecting more feedback on my implementation than the actual mechanics, that was kinda the main reason why i made the pr, but if noone cares and everyone hates the feature itself then so be it. ill wait for bob to reject/comment on it.

irislgtm:
>~< Come back to iq edition my liege

irislgtm:
Fuck markdown bro the fuck is that

HbmMods:
> because it looked more like a mistake than intentional design, since the durability setter was never actually changing the durability

Durability is set for most guns even ones that don't use it, since 0 values in a system where a lot of division is going on may not be a good idea to have. The fact that durability was never actually intended should have been obvious by the durability bar being absent from the HUD.

Also instead of making drill bits change the wear, they could have just changed the max durability instead, since that doesn't need a new weapon mod value that's not used anywhere else.

PR #3141, opened 2026-08-13
This PR will cause negativity again

TheVitya2127:
...

---

## Fixed
- Fixed a bug that caused the Lunar Miner to not require an Orbital Module
- Small GUI fix and small updated Locksmith Table GUI texture


RandomDude9587:
it barely even did anything, why would people cause negativity

TheVitya2127:
I don't know. People were really angry with the GUI and the new pylons. Now they blame all their problems on me.

PR #3144, opened 2026-08-14
Satellite ID Fix and I18nUtil (Slay)

TheVitya2127:
This poll request contains all changes related to satellites.

## Changed
- More translationification for satellites
- Minor changes in the Sescription of Sat Dock and Sat Linker
- Updated russian localization

I would also like to suggest new icons for info icons (it turns out they are very convenient. Why are they rarely used?)

изображение

---

For those who ran to write nasty things about me: I checked these innovations both on the client-side and on the server-side. Yup


WolfEclipses:
I mean, asking you to test something on the server when it explicitly says to isn't being mean or nasty. Description feels dramatic. Icons do match the current theme, in fact I don't know what the original vs new ones are, so you get feedback overall. Gray gear looks bland and dull, the blue gear looks a little too uh, saturated. Needs more of the m-word in the PR, very disappointing and disgruntling.

TheVitya2127:
> I mean, asking you to test something on the server when it explicitly says to isn't being mean or nasty.

It's not even about that, but about the situation as a whole.

> Description feels dramatic.

You're exaggerating.

>  Gray gear looks bland and dull, the blue gear looks a little too uh, saturated.

The palette can always be fixed. I just sent some rough sketches of how to improve it.

Orig:

изображение

---

> Needs more of the m-word in the PR, very disappointing and disgruntling.

A?

Randomdude0689:
vitya is on something, any guesses?

HbmMods:
> I would also like to suggest new icons for info icons

Looks neat, I wonder how well they fit with the new GUIs

>  (it turns out they are very convenient. Why are they rarely used?)

because i forgor

TheVitya2127:
> Looks neat, I wonder how well they fit with the new GUIs

@HbmMods 
изображение

изображение

изображение

PR #3148, opened 2026-08-16
Heating Oil QMAW

WolfEclipses:
Im not sure if this one is sufficient, but did not see any other uses that were, as common one might say.


PilotsMask:
could 'prolly mention that a common use for it is in fluid burners, since its well... a good heat source for them

PR #3149, opened 2026-08-16
New Electric Press Model

TheVitya2127:
The New VityaSlop™

изображение

---

- New Electric Press Model
  - By design, it is a repainted version of the Conveyor Press
  - The upper and lower energy ports are no longer working (slop)
  
изображение

I also really want to know your opinion about the palette. To create the palette, I created a mix between purple (my favorite color) and blue, so I will be glad to hear your opinion about this palette.

### Why is the new press model so weird?
Initially, I just wanted to make the model look similar to the original one by changing the port issues and texture resolution until I remembered the recipe for the conveyor press.

---

**It won't cause any problems with the update, I checked, honestly**


Randomdude0689:
I'd say it's better.

TheVitya2127:
Update: I just noticed a problem with the model. These gray frames overlap strangely with another block, so I suggest an alternative model.

изображение

---

изображение

TheVitya2127:
I'm not going to add a new model yet until I get an opinion about the current one (which is at the very top)

irisfalke:
its not very industrial which is whatever but i dont love the color/shading. i like the new alt model better out of the two

dandelion711:
I personally prefer the current press version over your design but thats just me

sten104:
this looks like shit, the one we have is infinitely better

TheVitya2127:
I can remake the model and make it similar in design to the original.

sten104:
> I can remake the model and make it similar in design to the original.

yes, perhaps you can add some pipes similar to those on the chemical plant and make the piston have a telescopic extension

InstantNootles:
ew

TheVitya2127:
> ew

Thanks for the very informative critique. It helped me understand what was wrong with the model, and now I'm going to make a new model mk.69

InstantNootles:
The electric press doesn't need a remodel, don't even start.

TheVitya2127:
> The electric press doesn't need a remodel, don't even start.

I did this solely because of port and texture resolution issues. Do people even read my PR description?!

InstantNootles:
If you did it solely because of the port and texture resolution issues then why did you completely change its design.

TheVitya2127:
> If you did it solely because of the port and texture resolution issues then why did you completely change its design.

*The reason for this decision is also written in the PR description*

hbmbals

I also suggested an option with the possibility of making a new model while preserving the original design.

InstantNootles:
You probably shouldn't just flat out copy the design of another machine and change it to some garish saturated color as a "redesign", that's just lazy and will confuse people.

TheVitya2127:
> You probably shouldn't just flat out copy the design of another machine and change it to some garish saturated color as a "redesign", that's just lazy and will confuse people.

***The reason for this design is also written in the PR description***

I just thought the conveyor press recipe was funny, so I decided to make it exactly as you see it now. Regarding the color, I initially asked for criticism due to issues with my monitor.

InstantNootles:
Yes, I know your reasons are written in the PR description, your reasons just suck. Quit saying that like it means anything.

TheVitya2127:
> Yes, I know your reasons are written in the PR description, your reasons just suck. Quit saying that like it means anything.

~~Your arguments suck too~~
However, I will now be working on creating a new model based on the original design.

HbmMods:
Personally not a fan, the thin supports on the conveyor press were only made this way in order to fit the conveyor belt under it without using up too much extra space, the electric press should have a thicker support.

Also the color gradient from purple to blue is a bit weird. The original blue color was supposed to look like crappy enamel, the gradient texture from the conveyor press doesn't really fit since that is supposed to represent unpainted copper/bronze.

TheVitya2127:
In that case, i'll remake the model

TheVitya2127:
*sounds of ass tearing*

Randomdude0689:
holy fucking shit go touch grass jesus

PR #3150, opened 2026-08-16
centrifuge valid input check

oqpr-ok:
changes the centrifuge so only items with centrifuge recipes can get inserted by a hopper/pneumatic/etc


RandomDude9587:
absolute win for bore setups

PR #3170, opened 2026-08-18
Additional rules for the .editorconfig

Dash1269:
This should bring the editor config to be largely inline with your coding style Bob.
I spoke with Martin about it and he said the config options look good so far.

This hopefully should help keep code in PRs more consistent and with nicer diffs for review (assuming the contributor runs their formatter)


HbmMods:
*starlight_glimmer_saying_balls_with_reverb.mp4*

PR #3179, opened 2026-08-19
Qmaw fixes

WolfEclipses:
Fixes some things like slight case issues in a couple files, reference issue in one, chlorine gas not using template reference for vacuum processing, also this updates the original oil 1 files I made way back when I started to the same style Im using now, this is partially a preference and partially since most were pre-style guide, they are all similar styles now.

-I also added pet gas to the oil 1 template, it felt weird not being there

-Tomorrow: I plan on adding a page for fuels and a page for diesel and light oil. this is because in the kerosene file I reference fuels, but I feel that should show a list of all fuels (the useful ones at least not just everything that can be burned). And I referenced diesel variants, well thats normal and cracked diesel, so I should make a page referencing both instead of just linking plain diesel.

Sorry for the yap, I dont normally talk in these so its kind of a info dump xd


src/main/resources/assets/hbm/manual/fluids/chlorine_gas.json
```diff
@@ -6,6 +6,6 @@
...
+        "en_US": "Early methods of getting chlorine gas include using glowstone or [[sodalite|Sodalite]] obtained from lapis ore in the [[combination oven|Combination Oven]]. Alternatively, it can be made with an [[electrolysis machine|Electrolysis Machine]] or gradually over time with the chlorine pinwheel. It has many uses including [[bedrock ore processing|Bedrock Ore Processing]] and producing [[high-performance solvent|High-Performance Solvent]]. One of its primary uses is the production of [[PVC]], or [[hard plastic|Hard Plastic]] through [[phosgene|Phosgene]]<br><br>{{Template:Oil3}}"
```
HbmMods: Does chlorine even need the oil 3 template? It's not an oil 3 product, and only related with the whole sector by being an ingredient for hardplastics.

WolfEclipses: I was 50/50 on that one originally, the ru-local people kept talking about it so I figured it must have been desired by some people, Ill change it back

WolfEclipses:
It has been fixed :P

PR #3192, opened 2026-08-21
Custom parts and mats

FOlkvangrField:
Add custom material and custom part system, just like the custom fluid system. Part textures and localized texts can be imported via resource packs.


TheVitya2127:
Why is it necessary for NTM except for individual Modpacks?

FOlkvangrField:
> Why is it necessary for NTM except for individual Modpacks?

Just like other custom systems in NTM such as custom fluids, custom machines and custom recipes, these features not only facilitate Modpacks in expanding NTM’s unique fluid and Mats systems, but also allow players to create their own items on their own, without having to go to GitHub to request the addition of numerous new items.

HbmMods:
Was this tested on a server?

FOlkvangrField:
> Was this tested on a server?

Now there should be no problem

PR #3199, opened 2026-08-22
Most useless fluid in the mod (Leaded Petroil)

WolfEclipses:
Besides water, like who needs that (joke)


sten104:
i think petroil could get a redemption if it had a good cracking recipe that uses it

WolfEclipses:
Simply, it needs more power considering the difficulty of its recipe, LPG for example only requires compressors and cracked oil, yet generates an insane amount of power, meanwhile this requires multiple automation processes that aren't as automatic, and generates far less power. However I also don't think it needs a buff, it's just, kind of there.

PR #3209, opened 2026-08-25
New Space Textures and I18nUtil

TheVitya2127:
It's been almost two months since my friend ~~fucked up~~ suggested that I make this PR with a new texture of a Photovoltaic Panel and an Orbital Module. Well, I finally got around to doing it.

изображение
изображение

There is also a minor GUI fix and a translation Soyuz-FG Desc.

изображение

---

*Of course, I checked everything and nothing broke on both sides.*


TheVitya2127:
I probably shouldn't have updated the Soyuz GUI

TheVitya2127:
I would also like to ask you a question about the skins for Soyuz. Will it be possible to find them through a special blueprint?

HbmMods:
Not a fan, the orbital module facing left instead of right made it face the same way as the soyuz and separates it from the satellites. I already have a replacement texture.

The photovoltaic panel's frame feels way too thick and the pattern's no longer the way I intended it to be.

> I would also like to ask you a question about the skins for Soyuz. Will it be possible to find them through a special blueprint?

The skins being obtainable is planned, but I haven't yet decided how that will work.

PR #3213, opened 2026-08-25
Fix tape drive shift-click item loss

Fammed-20:
Fixes tape drive shift-click allowing oversized drive stacks in single slots

### DIFF
--- a/src/main/java/com/hbm/inventory/container/ContainerTapeDrive.java
+++ b/src/main/java/com/hbm/inventory/container/ContainerTapeDrive.java
@@ -1,7 +1,12 @@
 package com.hbm.inventory.container;
 
+import com.hbm.util.InventoryUtil;
+
+import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.InventoryPlayer;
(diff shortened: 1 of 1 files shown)


HbmMods:
If I understand correctly, all this does is replacing the default merge behavior with `InventoryUtil.mergeItemStack`, which honestly should be the default behavior for all `ContainerBase` anyway.

Fammed-20:
> If I understand correctly, all this does is replacing the default merge behavior with `InventoryUtil.mergeItemStack`, which honestly should be the default behavior for all `ContainerBase` anyway.

@HbmMods 
Not quite, I recorded a video with the fix and without it. Please take a look.

without fix:

https://github.com/user-attachments/assets/daea6569-eb44-47cf-98df-d67494f16a0c

fix:

https://github.com/user-attachments/assets/325bd2bf-f75a-470f-9862-a45ed1d114a7

HbmMods:
Today's commit replaced the merge call in the container base class with `InventoryUtil.mergeItemStack` which indeed fixes the issue, not only for the tape drive but for any other potential container that uses the same base class.

PR #3215, opened 2026-08-26
QMAW Update I vague posted about (on accident)

WolfEclipses:
This adds:
-Basic tabbing functionality
-Bullet Point
-Centering

example of a file for it
```
{
    "name": "Reformate",
    "icon": ["hbm:item.fluid_icon", 1, 91],
    "trigger": [["hbm:item.fluid_icon", 1, 91]],
    "title": {
        "en_US": "Reformate"
    },
    "content": {
        "en_US": "Reformate can be <ct>obtained</ct><br> by [[vacuum refining|Vacuum Refinery]] [[pressurized|Compressor]] [[crude oil|Crude Oil]]. It has a diverse selection of uses including the production of <br><bt>[[reformate gas|Reformate Gas]] through <br><tb>[[cracking|Catalytic Cracking Tower]] or [[radiolysis|Radiolysis Chamber]]. Another notable use is the production of [[BTX]] through [[fractioning|Fractioning Tower]]. Additionally, it is a key ingredient in both [[high-cetane diesel|High-Cetane Diesel]] and [[high-cetane cracked diesel|High-Cetane Cracked Diesel]].<br><br>{{Template:Oil3}}"
    }
}
```
here is how it renders:

image

### DIFF
--- a/src/main/java/com/hbm/qmaw/GuiQMAW.java
+++ b/src/main/java/com/hbm/qmaw/GuiQMAW.java
@@ -14,6 +14,7 @@ import com.hbm.qmaw.components.*;
 import cpw.mods.fml.common.FMLCommonHandler;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.audio.PositionedSoundRecord;
+import net.minecraft.client.gui.FontRenderer;
 import net.minecraft.client.gui.GuiScreen;
 import net.minecraft.client.renderer.RenderHelper;
 import net.minecraft.client.resources.LanguageManager;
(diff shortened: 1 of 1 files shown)


HbmMods:
Honestly I'm not really sure how useful any of these are.

Centered text is only really usable for titles, and QMAW isn't really meant to have those at all, and with titles I always found that standard alignment is easier to read anyway. I can't really see any concrete use case where this would be necessary, as well as better than what we have already.

Bullet points can already be done with asterisks, tbh I find the high res unicode points to looks rather out of place.

Tabs are also something I'm not sure if we ever need that, considering that depending on what font renderer is used, the letter spacing is different anyway so having consistent indentation wouldn't really be possible. Proper indentation for bullet lists is nice, but it will break the instant the line width changes, either by the GUI itself changing (something I'm planning to do anyway) or once again by the font renderer being different.

WolfEclipses:
The centering is as you said primarily only good for titles, it won't do much on any specific page. I mainly meant for it on the oil templates, as even though it's green, making it standout more wouldn't hurt.

The bullet points I thought would be useful for specifically the crude, cracked, and vacuum oil pages where it has lists in the middle of the page. The templates would look weird like this, but those I think would look better. I was not aware it could be done with asterisks to be honest.

Tabbing is the most useless, it was intended as an alternative without bullet points, but other than that it is mostly useless.

Based on your feedback though bullet points and tabs I'll remove, I still think centering would be useful however

Edit: as for the centering never being fully accurate, yeah, I came to that conclusion and just kind of averaged it out at a 4px char average

Issue #3222, opened 2026-08-28
The satellites cooldown is being reset


TheVitya2127:
After the replay of the game, the Satellite Cooldown is reset.

Fammed-20:
Not quite, it's calculated based on real time

TheVitya2127:
Maybe it's just me, but after every restart of the game, the asteroid/lunar mining ships returned earlier than the specified time. Also, the Space Lab creates a data drive by default after 100 hours, and during all the restarts, 5 data drives appeared.

PR #3225, opened 2026-08-29
Fix satellite timer persistence

Fammed-20:
Prevents satellite timers and sensor progress from resetting after restarting the game. #3222


utafrali (approved):
The fix correctly identifies and patches the missing `markDirty()` calls after state mutations in all three satellite classes. The `sensorCount > 0` guard in `SatelliteScience.onUpdateTick` is a bonus cleanup that avoids redundant dirty-marking when no sensors are attached, since incrementing by zero is a no-op.

PR #3226, opened 2026-08-29
Add Brazilian Portuguese Localization

Karmatical:
**Brazilian Portuguese Localization Entry** (PT-BR) _(manuals are not complete yet, so for now i can only provide the .lang file)_

I also provided translation for items without reference (last section of the file), and legacy items.


HbmMods:
i have no means of verifying any of this but i'm gonna trust ralph deltarune

PR #3243, opened 2026-09-04
Fix satellite cargo delivery and display miner cooldowns

### LINKED ISSUE #3242: Hedgehog in Space
After the Cargo Landing Pad rework, the landing rockets began to behave strangely. I'm not sure if they're flying in principle, so I'd like to suggest adding a cooldown display to them like a space lab.

Image

Image

Fammed-20:
Preserves pending satellite cargo across world saves and cargo pod braking height across entity reloads. Cargo is now cleared only after the pod spawns successfully, and asteroid and lunar miner cooldowns are displayed through the Satellite Link overlay.

Fixes #3242


HbmMods:
Have to do these manually because the display I made in the prior commit is causing a merge conflict.

PR #3247, opened 2026-09-05
Fix satellite miner ready status and cargo pod collisions

### LINKED ISSUE #3242: Hedgehog in Space

Fammed-20:
Shows READY when asteroid or lunar miner cargo is waiting for delivery. Cargo pods now explode when colliding with blocks other than a Cargo Landing Pad instead of treating obstacles as valid landing surfaces.

Fixes #3242

### DIFF
--- a/src/main/java/com/hbm/entity/missile/EntitySatellitePod.java
+++ b/src/main/java/com/hbm/entity/missile/EntitySatellitePod.java
@@ -243,13 +243,11 @@ public class EntitySatellitePod extends EntityThrowableInterp implements IRadarD
 		
 		if(mop.typeOfHit == mop.typeOfHit.BLOCK) {
 			this.setPosition(mop.hitVec.xCoord, mop.hitVec.yCoord, mop.hitVec.zCoord);
-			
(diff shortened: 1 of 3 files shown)


HbmMods:
Cargo landing pods are no longer supposed to explode when landing on other blocks.

Fammed-20:
@HbmMods 
Oh, okay, I get it.
What about "Shows READY when asteroid or lunar miner cargo is waiting for delivery" instead of displaying 0%

HbmMods:
The "READY" part wasn't originally added because usually you'd just have a cargo landing pad linked anyway, which pulls the items almost instantly, so there's almost no window where you'd actually see it.

Fammed-20:
That makes sense, thanks for explaining! Without a linked pad, though, 0% can make it look like mining hasn't started. 
If you still feel this isn't needed, you can close the PR.

PR #3248, opened 2026-09-06
Satellite ID Manager automation

Fammed-20:
Allows compatible items to be inserted through pipes and conveyor inserters, with ID copying accessible from the sides and randomization from below. Prevents premature extraction and marks inventory and ID changes for saving.


HbmMods:
No longer needed with the latest commit, I did a full rewrite of the ID manager's tile class because it still implemented `ISidedInventory` directly instead of using one of the base classes. Automation of the randomizer slot is in my opinion not necessary, so only the copy slot can be accessed.

PR #3269, opened 2026-09-10
War crime or [250 CHOCOLATE]

TheVitya2127:
The Cargo Landing Pad recipe has been changed because... Why not?

изображение

изображение

---
The same reason.
Note: Since the military circuit is expensive and actually replaces the PCB, I changed the number of microchips from 8 to 4 compared to the sat version.

изображение

изображение

and I also changed the color of the display of player nicknames in the list of the turret chip.... Idk why

изображение


TheVitya2127:
Perhaps the turret chip should be moved to the assembly machine instead of my version.

TheVitya2127:
Also, maybe I added too much technetium steel.

PR #3273, opened 2026-09-11
Steams QMAW (Read description bob)

WolfEclipses:
Important to note: When viewing the page, there is no way of knowing that you can scroll down, its a long page, as there is so much important info about steam that the player really needs to know (cooling loop kind of deal).

Problem: When looking since you have no way of knowing that there is more text to the page, the scrollbar mainly is decoration on most pages (besides seeing Felix <3), I would like your opinion on some different texture or some method of knowing if there is more to the page so that people will actually scroll down. Anyway, hai bob hope you're having a good day! (rest is not important)

The steam page is finally here dear fellow daily PR enjoyers, on uhhh day 49, that means tomorrow will be a great day 50! Total qmaw page count: 83 fluids / 154 remaining.


HbmMods:
I'm planning on decoupling the QMAW GUI scale from the vanilla setting so you can actually read way more stuff than the 3 letters you get on auto scale, if that's not enough I could add like a scroll "page number" thing on the bottom.

PR #3290, opened 2026-09-14
Isnt this in my blood or smth (Plasma QMAW)

WolfEclipses:
This one links the deprecated plasmas, and through the fusion pages. Since the new plasmas arent fluids they arent linked. It mentions nothing about the old plasmas since they are yk, old, however since I assume you dont want an ID shift issue its better to just link them to information about the new shit


WolfEclipses:
73 remaining btw (This also means we are over halfway done with all ntm fluids!!!! Yippee!!!)

Randomdude0689:
I mean yes but also no. There is plasma in your blood, but not the kind used in fusion reactions.
Hope this helps!
Rate our AI feedback at [CENSORED].com/rate!

HbmMods:
I don't think it's a good idea to add this page to the *deprecated* plasma fluids which have nothing to do with the new fusion reactor. Nor does this page include that much information not already present in the fusion reactor ones.

PR #3291, opened 2026-09-14
First-class support for RoR/RTTY in opencomputers

Xpl0itR:
Adds an expansion card item that slots into a computer and presents as a component which can be used to either open handles to channels for sending and polling, or to bind OS events to be fired when the signal on a channel changes.


HbmMods:
Did you test this without OC installed? `Driver` extends `DriverItem` which is part of the OC API which the NTM jar doesn't include, so I think this may not actually work.

Xpl0itR:
So I did test it when I was registering the driver in my own func in MainRegistry, but then I came across an existing initialiser for OC so I moved mine there but didn't test again.

It works now.

PR #3294, opened 2026-09-15
fix hidden catalog hashing behavior

mlbv:
77f044ebdd13172de94646f7793cf89bcf31de0a changed the algorithm of ShadyUtil.smooth, making it impossible to pass the test; this patch reverts it

An alternative fix is to update the hash digests to `3d742d6b9303165191c3dedb86015365368e6a057bfd9f55435c33d487c9fcaa` and `8247c4cc73044f895f43fc7ee5e616434aa8de9ba388022bdeba332e225759b0`, matching the updated RNG behavior


NoErnesto:
afaik this change was intentional to prevent a certain person from wrecking all ntm servers in existence, cuz they were spamming hidden catalogues everywhere and couldn't keep the answers to themselves

HbmMods:
The smooshing function "works" as intended, the same way the Y/Z coordinate flip in the event handler is intended.

> afaik this change was intentional to prevent a certain person from wrecking all ntm servers in existence, cuz they were spamming hidden catalogues everywhere and couldn't keep the answers to themselves

It's not that likely that this was even true given how there have been no reports of the catalog being used anywhere since then.

PR #3298, opened 2026-09-15
Again Deco Block Addition 

TheMadLadRulz89:
Just the Checkerboard Tiles Block for Decorating

last time i messed something up with the code and i am sorry :(
checkerbord2
checkerbord1


TheMadLadRulz89:
i hope i didnt messed up anything this time
and i also added the Name in English and German

WolfEclipses:
This no longer causes problems, you might get rejected based on block ids not using meta, just a forewarning. It looks interesting though

TheMadLadRulz89:
> This no longer causes problems, you might get rejected based on block ids not using meta, just a forewarning. It looks interesting though

do i need to change something?
btw wouldnt that be too complicated to add MetaData to a simple Deco Block?

TheVitya2127:
Nuclear Furniture

HbmMods:
We are not doing any more deco blocks until the block ID situation has been resolved.

PR #3301, opened 2026-09-15
Electrolysis Apotheosis

TheVitya2127:
To be honest, in addition to the model, I would like to make configured ports right away, but due to the difficulties of the code, now the pull request contains only a model with a texture.

изображение

изображение

Sorry


WolfEclipses:
I do like this btw, it looks nice

TheVitya2127:
> I do like this btw, it looks nice

Sure, such a change of ports will break factories for many players, but I think that the price for a more convenient location of ports + 4 additional ones is justified

nuhyuh:
> > I do like this btw, it looks nice
> 
> Sure, such a change of ports will break factories for many players, but I think that the price for a more convenient location of ports + 4 additional ones is justified

honestly this is still a improvement

HbmMods:
Couldn't you have kept the original ports and just added the new ones in addition?

TheVitya2127:
> Couldn't you have kept the original ports and just added the new ones in addition?

It seems to me that it will look bad and not so harmonious, but I will do it.

TheVitya2127:
изображение
изображение

HbmMods:
grafik

PR #3306, opened 2026-09-16
creoseoteotes qmaw 👍 

WolfEclipses:
I misspell creosote as creosete every time for some reason


NoErnesto:
cereosteto 
FB_IMG_1784576161609

nuhyuh:
tetosphere

> cereosteto FB_IMG_1784576161609

TheVitya2127:
clueless_flix_hd

HbmMods:
"Burning" is weird phrasing given that it sounds like you're using the coal as fuel, even though you're coking it.

WolfEclipses:
> "Burning" is weird phrasing given that it sounds like you're using the coal as fuel, even though you're coking it.

is that what its called by putting it into a combination oven? weird... ok yeah ill change it, I was hoping burning would make more sense than some of the alternatives, but didnt know that coking was what it was called

PR #3308, opened 2026-09-16
AND Another Structure

TheMadLadRulz89:
Well i made a Military Base Structure that only spawns in Snowy biomes.
(its spawnweight is 1)
its basically a Dungeon with Traps and Enemies, but packed with bunch of Loot

Entrance of The Base

Back of the Base

interior of the radio and generator building

Barracks Interior with a Dud

The Storage Hall (yea i know it looks like there are no crates)

if it needs some tweaks, i'll work on it


WolfEclipses:
I like the concept, Im working on a hangar styled structure so uh, please avoid making one of those please and thank you 👍 

Otherwise yeah this looks cool

Randomdude0689:
The existence of a dud implies someone didn't do their job correctly

HbmMods:
Looks neat, how does it look on a non-flat world, though? The structure system tends to fall apart with large structures like that on uneven surfaces.

TheMadLadRulz89:
with some luck it could look like this


i also think i need to tweak it a bit so it has more ground

TheMadLadRulz89:
idk if i can send here an updated .nbt
nvm it did it form itself

TheMadLadRulz89:
it also could look like this

PR #3309, opened 2026-09-17
Add RoR commands to some machines

Xpl0itR:
RoR XD


TheVitya2127:
As far as I know, the ability to switch recipes using RoR for the Assembly and Chemical Factory was not added specifically.

HbmMods:
On the factories, the recipe switch was omitted on purpose.

PUREX and plasma forge never got recipe switching because those are so situational I can't really think of any reason why you would need that, plus most of the recipe variance can be handled with auto switch groups anyway.

Toggles for the large mining drill's configuration feel like a weird inclusion given that there's no indication as to what the drill is currently mining so you can't selectively enable silk touch or auto crusher based on that.

PR #3310, opened 2026-09-17
Hide RTTY card when OC is not loaded

### LINKED ISSUE #3296: What a...
After [this pull request](https://github.com/HbmMods/Hbm-s-Nuclear-Tech-GIT/pull/3291), a new card appears even without installing the OpenComputers mod.

Image

@Xpl0itR

Xpl0itR:
fixes #3296

also fix what I assume to be a mistake in a5f077a86f1c4ff07b6a81b64b6bd1210c22b252 as the build fails with the change


HbmMods:
The change in a5f077a86f1c4ff07b6a81b64b6bd1210c22b252 was done because on my second PC, the OC API seems to be different, even though the version number is identical, and I had to get the thing runnable somehow. Running the workspace setup/build didn't seem to make a difference.

Xpl0itR:
Hmm that's strange, maybe you have an older version stuck somehow. For clarity, the `*.network.EnvironmentHost` one is what is used on both the latest original and gtnh fork

https://github.com/MightyPirates/OpenComputers/blob/master-MC1.7.10/src/main/java/li/cil/oc/api/driver/Item.java#L65

https://github.com/GTNewHorizons/OpenComputers/blob/master/src/main/java/li/cil/oc/api/driver/Item.java#L65

PR #3318, opened 2026-09-18
Camp structures

TheMadLadRulz89:
I made some (again) Snowy Exclusive Structures, which are small Camps with two variations

Camp A with some Conserves and a Makeshift Gun Loot

Camp B as a Medical Tent

btw some of them look messed up because i couldnt find a good generated one but those should give players some useable stuff

and somehow my Contributor batch and my score disappeared :(


TheVitya2127:
It looks a lot like the camp from the new versions of Minecraft.

TheMadLadRulz89:
> It looks a lot like the camp from the new versions of Minecraft.

ok i didnt realise that, because i dont play the newer versions of the game

TheVitya2127:
> ok i didnt realise that, because i dont play the newer versions of the game

It's good. I don't play either)

HbmMods:
Idk, feels a bit bland

TheMadLadRulz89:
> Idk, feels a bit bland

i tought of that too tbh
it should be something small to find like the other small structures

PR #3330, opened 2026-09-19
Arty over radio

Xpl0itR:
Adds command to change targetting mode between auto and manual via RoR for the purpose of switching to manual mode to bombard a set of coordinates before switching back to auto mode.

Also changes the coordinate separator sent by the RoR designator from semicolon to the RoR variable separator to allow for easier processing on the autocal to send the coordinates to the artillery. This change is inconsequential to the use of the designator with satellites as they use a space as their separator so string manipulation would be required either way.


NULL-173:
I agree that this function is very lacking

PR #3332, opened 2026-09-20
Downloading... Autocal

EXPLOSIVEGAMER:
## Finally adds the Download option to the Autocal.
- Downloads the uploaded script of the autocal to the clients script.txt
- Testet with multiple clients on a server and in singleplayer.


WolfEclipses:
There is a reason this wasn't added. It was in place and got removed cause it's not safe yet.

EXPLOSIVEGAMER:
Ah ok. why is it not safe if i may ask?

legendarydoge30:
The fact that it could download any arbitrary text onto someone's computer if you were on a multiplayer server made bob omit the feature. It might be added in the future in some special way.

EXPLOSIVEGAMER:
Understandable. Did not think about that.

legendarydoge30:
I don't even know how dangerous a .txt file could be but thats the explanation I remember bob giving when he was adding the AUTOCAL.

RandomDude9587:
> I don't even know how dangerous a .txt file could be but thats the explanation I remember bob giving when he was adding the AUTOCAL.

that txt file can contain anything, which is a whole new batch of worms

EXPLOSIVEGAMER:
The file is set to be not executable but even with that it's still a blind download. I understand that if it gets added it has to have some sort of check to see if it is a autocal script.

PR #3336, opened 2026-09-21
Add item turret mob filter

NoPressF:
The item adds the ability to filter mobs using a blacklist or whitelist to determine which ones the turret should target. It also allows for importing and exporting the mob list

Item recipe:
image

Video demonstration of turret mob filter:
https://files.catbox.moe/hgr7hq.mp4


EXPLOSIVEGAMER:
I think it would be nice if you made the Item like an "Advanced Turret AI-Chip".
So it works the same as the Turret AI-Chip (So you have to physically put it in the turret)
For that you would also need to add players to your GUI.

Does the Chip also include filter options from the Machines list?

NoPressF:
No, the Chip doesn't include the Machines list

NoPressF:
Let’s imagine a situation where a player has 20 turrets and we know in advance that the filter will contain 10 mobs. It seems to me that crafting the same item every time would, firstly, not be particularly cheap, and secondly, it would be a faff to change the filter on 20 turrets, you’d literally have to change the list for 20 items and go to each turret. And if you needed to remove a single mob from the filter, you’d have to go round again and change the filter on all 20 turrets.

EXPLOSIVEGAMER:
You could work something out with the settings tool. And the recipe is not that expensive + It would be more logical to have it in the turret.

NoPressF:
I think there is so many stuff need to set up filter mob list

EXPLOSIVEGAMER:
> I think there is so many stuff need to set up filter mob list

What precisely do you mean?

NoPressF:
I mean that you need 2 items to configure mob filter list. It's a little bit complicated

EXPLOSIVEGAMER:
Ah. I meant that you extend your own item with ItemTurretBiometry and use your item in the chip slot. But for that to work you have to re write quit a bit.

PR #3337, opened 2026-09-21
New CIWS Ammo Recipe

EXPLOSIVEGAMER:
Switched the Goalkeeper Twin CIWS 200 Round Belt recipe to the Ammo Press and removed Crafting Table recipe.

image


TheVitya2127:
Maybe replace copper with gunmetal?

EXPLOSIVEGAMER:
Great idea. Makes sense sinces moast ammo uses it. Going to change it tomorrow.

HbmMods:
Already took care of it, there were several other changes to the ammo belt necessary as well.

As for the recipe itself:
* The belt has to work against missiles so technically it would fall under FMJ (or even AP) which means the bullet needs to be steel
* The casings should at the very least be plates
* Due to the way automation works, identical item types should not be spread out over the grid as they are on the crafting table, otherwise automation would have to fill up two of the three slots completely before the third one can be accessed
* The price for the belt was lower than intended due to constraints with the crafting table, I increased material cost from 3 to 5 per item

